PR #3, opened 2022-05-05
Start WP in setup script.

StevenDufresne:
`package.json` is missing `wp-env start` from its scripts and if you hit a dead-end following the readme.


ryelle:
I just added the "start" command as a separate step in the setup process, does that work? I'd like to keep the setup script as just setup, but I don't feel strongly about that if you think combining them is better.

StevenDufresne:
Yep, that works too.

PR #6, opened 2022-05-11
Tools: Add workflow to build theme

### LINKED ISSUE #5: Create a automatically generated build branch
See https://github.com/WordPress/wporg-mu-plugins/pull/175

If we want to use this project in separate child-theme repos, it needs to have a branch with the built CSS files.

ryelle:
Fixes #5 — Create a build branch that only contains the parent theme files, along with a new `composer.json` to identify this as a `wordpress-theme`.

You can verify it's working by looking at [the Checks tab](https://github.com/WordPress/wporg-parent-2021/pull/6/checks) & [the build branch](https://github.com/WordPress/wporg-parent-2021/tree/build) generated by this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,15 +26,26 @@ jobs:
             - name: Build
               run: yarn workspaces run build
 
-            - name: Ignore .gitignore
+            - name: Trim the repo down to just the theme
               run: |
-                  git add source/wp-content/themes/wporg-parent-2021/build/* --force
+                  cp -R source/wp-content/themes/wporg-parent-2021 theme
+                  git rm -r .
+                  mv theme/ .
+                  rm -r theme
+
+            # Add a composer.json so we can set the type to `wordpress-theme`.
+            - name: Create the composer.json
+              run: |
+                  echo "{\n\t\"name\": \"wporg/wporg-parent-2021\",\n\t\"type\": \"wordpress-theme\"\n}" > composer.json
+
+            - name: Add all the theme files
+              run: |
+                  git add * --force
 
             - name: Commit and push
               uses: actions-js/push@a52398fac807b0c1e5f1492c969b477c8560a0ba # 1.3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+            - name: Build
+              run: yarn workspaces run build
+
+            - name: Trim the repo down to just the theme
+              run: |
+                  cp -R source/wp-content/themes/wporg-parent-2021 theme
+                  rm -rf theme/node_modules
+                  git rm -r .
+                  mv theme/* .
+                  rm -r theme
```
dd32: Looking at https://github.com/WordPress/wporg-parent-2021/tree/build I can see a `node_modules`, `vendor` and `source/wp-content` folders still.

This seems to clear it up entirely though to just the contents of the wporg-parent-2021:
```suggestion
            - name: Trim the repo down to just the theme
              run: |
                  rm -rf source/wp-content/themes/wporg-parent-2021/node_modules
                  mv source/wp-content/themes/wporg-parent-2021 $RUNNER_TEMP
                  git rm -rfq .
                  rm -rf *
                  mv $RUNNER_TEMP/wporg-parent-2021/* .
```

I rm'd the node_modules from the theme folder first, to reduce the time taken to move the directory. Minor optimization, but also for clarity that we're dropping the themes `node_modules` folder too.

By using `$RUNNER_TEMP` we can move the files we want to keep outside of the checkout. RUNNER_TEMP is cleared before/after each task.

By using `rm` after `git rm` we can remove the non-tracked files (eg node_modules and vendor). I'm not sure why `source/wp-content` was remaining, but this cleans that up too.

`git rm -q` is just to silence the list of files it's RM'ing, since that's irrelevant in the job logs and was confusing.

ryelle: Thanks! I've applied your changes and it looks better here too 💯

PR #16, opened 2022-07-28
Pattern: 2 column with CTA

### LINKED ISSUE #14: Pattern: Two column CTA
image

adamwoodnz:
See #14

Initial use of this pattern is on the downloads page:

Screen Shot 2022-07-29 at 8 39 26 AM

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Title: 2 Columns with CTA
+ * Slug: wporg-parent-2021/two-columns-with-cta
+ * Categories: columns, buttons
+ */
+
+?>
+<!-- wp:columns {"align":"full","style":{"border":{"top":{"color":"var:preset|color|light-grey"}}},"className":"is-style-default"} -->
+<div class="wp-block-columns alignfull is-style-default" style="border-top-color:var(--wp--preset--color--light-grey)"><!-- wp:column {"width":"100%","style":{"spacing":{"padding":{"top":"80px","right":"100px","bottom":"100px","left":"100px"}}}} -->
+<div class="wp-block-column" style="padding-top:80px;padding-right:100px;padding-bottom:100px;padding-left:100px;flex-basis:100%"><!-- wp:heading {"level":4} -->
+<h4>Download and install yourself</h4>
+<!-- /wp:heading -->
+
+<!-- wp:paragraph -->
+<p>For anyone comfortable getting their own hosting and domain.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:buttons -->
+<div class="wp-block-buttons"><!-- wp:button {"style":{"spacing":{"padding":{"right":"32px","left":"32px"}}}} -->
+<div class="wp-block-button"><a class="wp-block-button__link wp-element-button" style="padding-right:32px;padding-left:32px">Download WordPress 6.0</a></div>
+<!-- /wp:button -->
+
+<!-- wp:button {"textColor":"blue-1","style":{"spacing":{"padding":{"left":"32px","right":"32px"}}},"className":"is-style-outline"} -->
+<div class="wp-block-button is-style-outline"><a class="wp-block-button__link has-blue-1-color has-text-color wp-element-button" style="padding-right:32px;padding-left:32px">Installation guide</a></div>
+<!-- /wp:button --></div>
+<!-- /wp:buttons -->
+
+<!-- wp:paragraph {"textColor":"dark-strokes-grey","fontSize":"tiny"} -->
+<p class="has-dark-strokes-grey-color has-text-color has-tiny-font-size">Recommend PHP 7.4 or greater and MySQL 5.7 or MariaDB version 10.3 or greater.</p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:column -->
+
+<!-- wp:column {"width":"100%","style":{"spacing":{"padding":{"top":"80px","right":"100px","bottom":"80px","left":"100px"}},"border":{"left":{"color":"var:preset|color|light-grey"}}}} -->
+<div class="wp-block-column" style="border-left-color:var(--wp--preset--color--light-grey);padding-top:80px;padding-right:100px;padding-bottom:80px;padding-left:100px;flex-basis:100%"><!-- wp:heading {"level":4} -->
+<h4>Set up with a hosting provider</h4>
+<!-- /wp:heading -->
+
+<!-- wp:paragraph -->
+<p>For anyone looking for the simplest way to start.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:buttons -->
+<div class="wp-block-buttons"><!-- wp:button {"style":{"spacing":{"padding":{"left":"32px","right":"32px"}}}} -->
+<div class="wp-block-button"><a class="wp-block-button__link wp-element-button" style="padding-right:32px;padding-left:32px">See all recommended hosts</a></div>
+<!-- /wp:button --></div>
+<!-- /wp:buttons --></div>
+<!-- /wp:column --></div>
+<!-- /wp:columns -->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
@@ -6,44 +6,64 @@
  */
 
 ?>
-<!-- wp:columns {"align":"full","style":{"border":{"top":{"color":"var:preset|color|light-grey"}}},"className":"is-style-default"} -->
-<div class="wp-block-columns alignfull is-style-default" style="border-top-color:var(--wp--preset--color--light-grey)"><!-- wp:column {"width":"100%","style":{"spacing":{"padding":{"top":"80px","right":"100px","bottom":"100px","left":"100px"}}}} -->
-<div class="wp-block-column" style="padding-top:80px;padding-right:100px;padding-bottom:100px;padding-left:100px;flex-basis:100%"><!-- wp:heading {"level":4} -->
-<h4>Download and install yourself</h4>
-<!-- /wp:heading -->
 
-<!-- wp:paragraph -->
-<p>For anyone comfortable getting their own hosting and domain.</p>
-<!-- /wp:paragraph -->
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/**
+ * Title: 2 Columns with CTA
+ * Slug: wporg-parent-2021/two-columns-with-cta
+ * Categories: columns, buttons
+ */
+
+?>
+<!-- wp:columns {"align":"full","style":{"border":{"top":{"color":"var:preset|color|light-grey"}}},"className":"is-style-default"} -->
+<div class="wp-block-columns alignfull is-style-default" style="border-top-color:var(--wp--preset--color--light-grey)"><!-- wp:column {"width":"100%","style":{"spacing":{"padding":{"top":"80px","right":"100px","bottom":"100px","left":"100px"}}}} -->
```
adamwoodnz: This border color doesn't match the design but I've assumed we will adjust this in the theme settings?

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
...
+<div class="wp-block-column" style="padding-top:80px;padding-right:100px;padding-bottom:100px;padding-left:100px;flex-basis:100%"><!-- wp:heading {"level":4} -->
```
adamwoodnz: Padding needs responsive overrides on mobile

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
...
+<h4>Download and install yourself</h4>
+<!-- /wp:heading -->
+
+<!-- wp:paragraph -->
+<p>For anyone comfortable getting their own hosting and domain.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:buttons -->
+<div class="wp-block-buttons"><!-- wp:button {"style":{"spacing":{"padding":{"right":"32px","left":"32px"}}}} -->
+<div class="wp-block-button"><a class="wp-block-button__link wp-element-button" style="padding-right:32px;padding-left:32px">Download WordPress 6.0</a></div>
```
adamwoodnz: Default Button background color doesn't match designs. Not sure whether we would adjust button padding in theme settings too.

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
...
+<!-- /wp:button -->
+
+<!-- wp:button {"textColor":"blue-1","style":{"spacing":{"padding":{"left":"32px","right":"32px"}}},"className":"is-style-outline"} -->
+<div class="wp-block-button is-style-outline"><a class="wp-block-button__link has-blue-1-color has-text-color wp-element-button" style="padding-right:32px;padding-left:32px">Installation guide</a></div>
```
adamwoodnz: Color selection for outline and text matches default button background color

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
...
+<!-- /wp:button --></div>
+<!-- /wp:buttons -->
+
+<!-- wp:paragraph {"textColor":"dark-strokes-grey","fontSize":"tiny"} -->
+<p class="has-dark-strokes-grey-color has-text-color has-tiny-font-size">Recommend PHP 7.4 or greater and MySQL 5.7 or MariaDB version 10.3 or greater.</p>
```
adamwoodnz: Text color and size doesn't match designs

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
```diff
@@ -0,0 +1,49 @@
...
+<!-- /wp:paragraph --></div>
+<!-- /wp:column -->
+
+<!-- wp:column {"width":"100%","style":{"spacing":{"padding":{"top":"80px","right":"100px","bottom":"80px","left":"100px"}},"border":{"left":{"color":"var:preset|color|light-grey"}}}} -->
+<div class="wp-block-column" style="border-left-color:var(--wp--preset--color--light-grey);padding-top:80px;padding-right:100px;padding-bottom:80px;padding-left:100px;flex-basis:100%"><!-- wp:heading {"level":4} -->
```
adamwoodnz: This border style will need to be switched from left to top on small screens

adamwoodnz: Also this border isn't truly centered on the screen due to the column gap

Screen Shot 2022-07-29 at 8 52 04 AM

We'd need CSS to correct this, or adjust via theme/block settings maybe?

StevenDufresne: If we wanted to keep this within the editor, you can create a column in between the two and set its width to `1px` with the background color you want. With the current Gutenberg style, it would disappear on mobile and probably be close enough to the design. For me, the visual divider on mobile would be distracting.

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
StevenDufresne: Is there a way for us to use `REM` units for all the copy/heading and then utilize `em` units for padding? Extend the column block to allow us to set font-sizes for different breakpoints to the parent container.

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
tellyworth: I don't know if this matters, but I get block errors when inserting this into a post:
Screen Shot 2022-07-29 at 3 12 19 pm

adamwoodnz: @tellyworth I can't repro this but I have seen this error before during my support rotation and I'm sure there was an existing issue for it... I'll try to find it

StevenDufresne: I did not run into that issue when I was testing.

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
ryelle: You can use the padding preset properties, which we can update to use clamp or have responsive changes in the CSS. [See `parts/cover.html` for an example](https://github.com/WordPress/wporg-parent-2021/blob/7acf81d3297b4327177cb7e5df885269c538280a/source/wp-content/themes/wporg-parent-2021/parts/cover.html#L1-L2), but the attribute uses his format: `"padding":{"top":"var:preset|spacing|60",…`, and then the inline CSS uses the property, `padding-top:var(--wp--preset--spacing--60)`.

I listed out some of the defined properties here: https://github.com/WordPress/wporg-parent-2021/pull/18#issuecomment-1199977183

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
ryelle: I updated the button styles with #18, so the default styles should match now - let me know if it's not right, though 🙂

source/wp-content/themes/wporg-parent-2021/patterns/two-columns-with-cta.php
adamwoodnz: Thanks for this. I've nearly got it working but for some reason when I insert the pattern the columns are full width rather than wide, even though the class is in the markup:

```
<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":"0px"}},"className":"two-columns-cta"} -->
<div class="wp-block-columns alignwide two-columns-cta">
```

I can't for the life of me work out how to set this... as you can see there is no alignment setting on the block options in the editor. How did you manage to set up this layout?

Screen Shot 2022-08-02 at 4 58 34 PM

ryelle: Check the group block that's wrapping the columns - is the Layout toggle on? If so, it overrides the alignment of child elements & forces them to be the full width of the group. If you untoggle that, the alignment options on the Columns should come back.

Screen Shot 2022-08-02 at 10 16 03 AM

adamwoodnz: Ah that was it, thanks!

adamwoodnz: Updated to use the Large padding with clamping, works quite well

source/wp-content/themes/wporg-parent-2021/theme.json
```diff
@@ -340,7 +340,7 @@
 			},
 			"layout": {
 				"contentSize": "680px",
-				"wideSize": "1160px",
+				"wideSize": "1280px",
```
adamwoodnz: This is the minimum width required for the buttons to be able to sit next to each other on desktop, with the current padding of the buttons and columns. It may have wider effects we're not happy with though.

Screen Shot 2022-08-03 at 12 11 05 PM

source/wp-content/themes/wporg-parent-2021/theme.json
```diff
@@ -505,7 +505,7 @@
 			"fontSizes": [
 				{
 					"name": "Extra Small",
-					"size": "12px",
+					"size": "11px",
```
adamwoodnz: This text is specified as 11px

Screen Shot 2022-08-03 at 12 08 01 PM

source/wp-content/themes/wporg-parent-2021/theme.json
ryelle: It's because this text has the `6.0.1` while the mockup has `6.0` — leave this size as-is, and we can play with the padding and such on the site itself. The important part of this pattern is the two column layout, the side-by-side buttons can be managed once the pattern is used.

source/wp-content/themes/wporg-parent-2021/theme.json
ryelle: The `12px` came from the style guide - I think you can either leave this text as 12px, or set it to a custom size of 11px.

PR #19, opened 2022-08-01
Pattern: Add heading with arrow

### LINKED ISSUE #9: Pattern: Heading with arrow
This pattern is like the "Latest release" section on [News](https://wordpress.org/news/). A big heading, with an arrow, in a container with a background color.

Mockup:

image

The container group might also have other content (or maybe this is unrelated content just below it?)

Screen Shot 2022-07-20 at 4 59 13 PM

Screen Shot 2022-07-20 at 5 00 52 PM

outdoor2kode:
See #9.

### Screencasts

https://user-images.githubusercontent.com/18050944/182235432-0f862931-904e-4ac5-a0d2-febd6726e617.mp4

https://user-images.githubusercontent.com/18050944/182235452-ef943408-b6f0-4fae-a49a-c1a34ae1c9a2.mp4

https://user-images.githubusercontent.com/18050944/182235457-f5e9af72-54f2-4d80-9e5e-d6f6990f87dc.mp4

1. Check out to this branch.
2. Execute `composer update`.
3. Test patterns in post editor or site editor.

### TODO

Since I'm not sure if I've done things in a right way or not so far, I'll work on the responsive stuff after I get some feedback.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -192,7 +192,6 @@
 		display: flex;
 		flex-direction: column;
 		align-items: left;
-		font-size: var(--wp--preset--font-size--heading-cta);
 		text-decoration: none;
 
 		&::after {
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -175,3 +175,71 @@
 	font-size: var(--wp--preset--font-size--heading-5);
 	line-height: var(--wp--custom--heading--level-5--typography--line-height);
 }
+
+.wporg-pattern__heading-with-arrow {
+	display: flex;
+	flex-direction: row;
+	justify-content: center;
+
+	&:hover,
+	&:focus-within {
+		background-color: rgba(0, 0, 0, 0.1);
+		transition: background-color 0.15s linear;
+	}
+
+	a {
+		display: flex;
+		flex-direction: column;
+		align-items: left;
+		font-size: var(--wp--preset--font-size--heading-2);
+
+		&::after {
+			content: "";
+			display: block;
+			height: 38px;
+			background-image: url(../images/right-arrow.svg);
+			background-position: center right;
+			background-repeat: no-repeat;
+			pointer-events: none;
+			transition: all 0.15s linear;
+			margin-top: 10px;
+			width: 88vw;
+		}
+
+		&:hover,
+		&:focus-within {
+			&::after {
+				transform: translateX(10px);
+			}
+		}
+
+		@include break-small() {
+			font-size: var(--wp--preset--font-size--heading-1);
+
+			&::after {
+				width: max(62vw, 100%);
+			}
+		}
+
+		@include break-xlarge() {
+			font-size: var(--wp--preset--font-size--heading-cta);
+			flex-direction: row;
+			align-items: center;
+
+			&::after {
+				height: 118px;
+				width: 35vw;
+				margin-left: 40px;
+				margin-top: 0;
+				transform: scale(0.9);
+			}
+
+			&:hover,
+			&:focus-within {
+				&::after {
+					transform: scale(0.9) translateX(10px);
+				}
+			}
+		}
+	}
+}
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -192,7 +192,7 @@
 		display: flex;
 		flex-direction: column;
 		align-items: left;
-		font-size: var(--wp--preset--font-size--heading-2);
+		font-size: var(--wp--preset--font-size--heading-cta);
 		text-decoration: none;
 
 		&::after {
@@ -220,15 +220,12 @@
 		}
 
 		@include break-small() {
-			font-size: var(--wp--preset--font-size--heading-1);
-
 			&::after {
 				width: max(62vw, 100%);
 			}
 		}
 
 		@include break-xlarge() {
-			font-size: var(--wp--preset--font-size--heading-cta);
 			flex-direction: row;
 			align-items: center;
 
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -187,11 +187,13 @@
 		transition: background-color 0.15s linear;
 	}
 
-	a {
+	// href added for specificity.
+	a[href] {
 		display: flex;
 		flex-direction: column;
 		align-items: left;
 		font-size: var(--wp--preset--font-size--heading-2);
+		text-decoration: none;
 
 		&::after {
 			content: "";
@@ -207,7 +209,11 @@
 		}
 
 		&:hover,
-		&:focus-within {
+		&:focus {
+			text-decoration-line: underline;
+			text-decoration-thickness: 1px;
+			text-underline-offset: 0.15em;
+
 			&::after {
 				transform: translateX(10px);
 			}
--- a/source/wp-content/themes/wporg-parent-2021/images/right-arrow.svg
+++ b/source/wp-content/themes/wporg-parent-2021/images/right-arrow.svg
@@ -1,0 +1,1 @@
+<svg width="670" height="120" fill="none" xmlns="http://www.w3.org/2000/svg"><path stroke="#fff" stroke-width="1.5" d="M0 60h669M609.928 1l59 59-59 59"/></svg>
--- a/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * Title: Heading with arrow
+ * Slug: wporg-parent-2021/heading-with-arrow
+ * Categories: header
+ */
+
+?>
+<!-- wp:group {"align":"full","backgroundColor":"blueberry-1"} -->
+<div class="wp-block-group alignfull has-blueberry-1-background-color has-background">
+	<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"97px","bottom":"97px"},"blockGap":"0px"}},"className":"wporg-pattern__heading-with-arrow"} -->
+	<div class="wp-block-group alignfull wporg-pattern__heading-with-arrow has-link-color" style="padding-top:97px;padding-bottom:97px">
+		<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"fontFamily":"inter"} -->
+		<h2 class="has-inter-font-family" style="font-style:normal;font-weight:300;letter-spacing:-0.02em;line-height:1.1"><a href="https://wordpress.org/download/">Get started</a></h2>
+		<!-- /wp:heading -->
+	</div>
+	<!-- /wp:group -->
+</div>
+<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
@@ -10,7 +10,7 @@
 <div class="wp-block-group alignfull has-blueberry-1-background-color has-background">
 	<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"97px","bottom":"97px"},"blockGap":"0px"}},"className":"wporg-pattern__heading-with-arrow"} -->
 	<div class="wp-block-group alignfull wporg-pattern__heading-with-arrow has-link-color" style="padding-top:97px;padding-bottom:97px">
-		<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"fontFamily":"inter"} -->
+		<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"fontSize":"heading-cta","fontFamily":"inter"} -->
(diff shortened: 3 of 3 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+            transform: translateX(10px);
+			transition: all 0.15s linear;
+        }
+
+		.heading-with-arrow__row > div > p {
+			text-decoration: underline;
+			text-decoration-thickness: 3px;
+			text-underline-offset: 15px;
+		}
+    }
```
adamwoodnz, line 154: Indentation is a little off through here

adamwoodnz: Indentation of some lines is off

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
```diff
@@ -0,0 +1,43 @@
...
+				<!-- wp:group {"layout":{"contentSize":"760px","inherit":false}} -->
+				<div class="wp-block-group">
+					<!-- wp:paragraph {"style":{"typography":{"fontSize":"120px","fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"textColor":"white"} -->
+					<p class="has-white-color has-text-color" style="font-size:120px;font-style:normal;font-weight:300;letter-spacing:-0.02em;line-height:1.1">We'll help you get started</p>
+					<!-- /wp:paragraph -->
+				</div>
+				<!-- /wp:group -->
+
+				<!-- wp:image {"id":8960,"width":370,"height":120,"sizeSlug":"full","linkDestination":"none","style":{"spacing":{"margin":{"top":"0px","right":"0px","bottom":"0px","left":"0px"}}}} -->
+				<figure class="wp-block-image size-full is-resized" style="margin-top:0px;margin-right:0px;margin-bottom:0px;margin-left:0px"><img src="https://wordpress.org/main-test/files/2022/07/Arrow-3.png" alt="" class="wp-image-8960" width="370" height="120"/></figure>
```
adamwoodnz: I think this would be better as a background image rather than content. News uses a pseudo element for this:

Screen Shot 2022-08-02 at 1 43 37 PM

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
outdoor2kode: Thanks for catching this. Fixed.

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
outdoor2kode: The reason I made it content was that I thought there was a possibility that users would want to change the image using the GB UI tools, and it feels like it is more flexible as a pattern for users to use. Not so sure about this, @ryelle what do you think about this?

outdoor2kode:
@ryelle I'm wondering do we need to make a pattern for
image
Or it's enough for us to provide the patterns that are created in this PR and let users tweak it themselves to the above style?

And the same question for the button animation style, do we tend to let users add it by themselves or should I refer to the `Get WordPress` button style in w.org and add it directly?

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
ryelle: No, the arrow doesn't need to be editable. You can use the same method as News.

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
```diff
@@ -0,0 +1,43 @@
...
+<!-- wp:group {"align":"full","backgroundColor":"blueberry-1"} -->
+<div class="wp-block-group alignfull has-blueberry-1-background-color has-background">
+	<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"97px","bottom":"97px","right":"80px","left":"80px"}}},"className":"heading-with-arrow","layout":{"type":"flex","orientation":"vertical","justifyContent":"center","flexWrap":"nowrap"}} -->
+	<div class="wp-block-group alignfull heading-with-arrow" style="padding-top:97px;padding-right:80px;padding-bottom:97px;padding-left:80px">
+		<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}},"layout":{"inherit":false}} -->
+		<div class="wp-block-group">
+			<!-- wp:group {"style":{"spacing":{"padding":{"bottom":"67px"}}},"className":"heading-with-arrow__row","layout":{"type":"flex","flexWrap":"nowrap","orientation":"horizontal","justifyContent":"center","verticalAlignment":"top"}} -->
+			<div class="wp-block-group heading-with-arrow__row" style="padding-bottom:67px">
+				<!-- wp:group {"layout":{"contentSize":"760px","inherit":false}} -->
+				<div class="wp-block-group">
```
ryelle: Why do you need this many group blocks? Could it be simplified?

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+// A heading with an arrow and other content.
+.heading-with-arrow {
+	transition: background-color 0.15s linear;
+
+	figure {
+		transition: all 0.15s linear;
+	}
+
+	a {
+		color: var(--wp--preset--color--white) !important;
```
ryelle: Rather than setting this here, can you set the link color on the parent group block?

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+		text-underline-offset: 2px;
+		cursor: pointer;
```
ryelle: Are these necessary? If we want to offset the underline, we should probably do that for all links.

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
```diff
@@ -0,0 +1,43 @@
...
+					<!-- wp:paragraph {"style":{"typography":{"fontSize":"120px","fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"textColor":"white"} -->
+					<p class="has-white-color has-text-color" style="font-size:120px;font-style:normal;font-weight:300;letter-spacing:-0.02em;line-height:1.1">We'll help you get started</p>
```
ryelle: This should be a heading block (you can default to `h2`), and the text inside it should be a link.

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+		.heading-with-arrow__row > div > p {
+			text-decoration: underline;
+			text-decoration-thickness: 3px;
+			text-underline-offset: 15px;
+		}
+	}
+
+	.heading-with-arrow__row {
+		cursor: pointer;
+	}
```
ryelle: This isn't an interactive element, so it shouldn't have a cursor style.

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+		figure {
+			transform: translateX(10px);
+			transition: all 0.15s linear;
+		}
+
+		.heading-with-arrow__row > div > p {
+			text-decoration: underline;
+			text-decoration-thickness: 3px;
+			text-underline-offset: 15px;
+		}
```
ryelle: Once the pattern is updated to use a link here, this shouldn't be necessary

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
 		}
 	}
 }
+
+// A heading with an arrow and other content.
+.heading-with-arrow {
```
ryelle: This should probably have some kind of prefix… maybe `wporg-pattern__`?

```suggestion
.wporg-pattern__heading-with-arrow {
```

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-list.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * Title: Heading with arrow and list
+ * Slug: wporg-parent-2021/heading-with-arrow-and-list
+ * Categories: header
+ */
```
ryelle: I'm not sure if we need 3 variations of this pattern, the original issue was just to create the heading & arrow. I don't feel strongly about that though, so if you think all 3 are useful we can keep them all in.

My comments about the block structure apply to all 3.

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,51 @@
...
+			text-decoration-thickness: 3px;
+			text-underline-offset: 15px;
+		}
+	}
+
+	.heading-with-arrow__row {
+		cursor: pointer;
+	}
+
+	.heading-with-arrow__list {
```
ryelle: We can switch this list to use [the list of links block style](https://github.com/WordPress/wporg-main-2022/issues/29) once it exists, rather than adding CSS here. (just noting that, no specific change requested now)

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-button.php
outdoor2kode: Originally I was thinking to make the patterns able to be customized easily with GB UI tools from users. So I tried to make the pattern with GB UI tools as much as possible. In this example, basically every aspect of the UI layout can be tweaked through GB UI tools. But from other comments, I guess I know what we're trying to make here. I'll adjust it!

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
outdoor2kode: I was trying to follow the design on figma as much as possible. But I guess you're right, this should be done for all links if we want it. I'll remove the style first as if applying it to all links, it probably would make the style go off for some fonts.
And ah, I noticed that in editor `<a>` tag would not display the pointer cursor, it would show up when it is published, will remove the cursor one as well.

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
outdoor2kode: I've been thinking about what prefix to use here... Thanks!

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow-and-list.php
outdoor2kode: I'm curious who would eventually use these patterns?
Would the users be ourselves or other developers only? If it would be used by some non-tech guys, I'm thinking that providing the 3 variations would still save much time for them.

ryelle: They'll be used by anyone editing the site - probably designers & writers, but also our team too. I think most people using them will have a good amount of gutenberg experience.

outdoor2kode: Cool, I'll remove them.

outdoor2kode:
New commit pushed and screencast updated.

https://user-images.githubusercontent.com/18050944/182952396-8b5e81bf-772a-41bc-bf27-f15de4684056.mp4

Currently there are three issues:

1. The underline thickness of the heading is thicker in the editor than published post.
2. The underline of the heading would disappear when hovering on it and show up when doing nothing, which is different from the one on News.
3. In Mobile Preview, font-size would be correctly set to 50px in the editor, but wrongly set to 30px when published.

I'm thinking that these are just local env issues or I missed some steps when publishing it. Not sure if I'm right about it.

ryelle:
> The underline thickness of the heading is thicker in the editor than published post.
> The underline of the heading would disappear when hovering on it and show up when doing nothing, which is different from the one on News.

Looking at the design, this shouldn't have an underline. It should act more like News, with no underline by default, and add the underline on hover. I've done that in https://github.com/WordPress/wporg-parent-2021/pull/19/commits/c8bb00d649dc8038aafb27baefb734e99c850a8f.

> font-size would be correctly set to 50px in the editor, but wrongly set to 30px when published.

You can just use the CTA heading size, it automatically adjusts down to 52px on small screens. I've made that change in https://github.com/WordPress/wporg-parent-2021/pull/19/commits/47c579ada39b5336ce919dee4311de51e1835849. Actually, as I'm writing this up, I realized we can just set that on the heading itself, so I've done that in https://github.com/WordPress/wporg-parent-2021/pull/19/commits/49c061aa4c771a29a511f1a9f304abf0c9fc510b.

I think this is good to merge now. There's something awkward happening at mid-sized screens (600 - 1080), but we can figure that out in a later PR.

Screen Shot 2022-08-04 at 6 31 15 PM

outdoor2kode:
> Looking at the design, this shouldn't have an underline. It should act more like News, with no underline by default, and add the underline on hover. I've done that in [c8bb00d](https://github.com/WordPress/wporg-parent-2021/commit/c8bb00d649dc8038aafb27baefb734e99c850a8f).

Originally I was thinking if it's something related to env provision because I can see that there are no explicitly underline styles set for News, instead I saw it has this parent style. Haven't had time to look into it and figure out how it works at the moment, so I'm not so sure about it.

image

But I guess it's also ok to set it explicitly here, originally I got another concern that it may affect the links that are added in the future to this group (normal links should have a default style like image below shows), but I think we can just add a new paragraph next to the group, instead of inside the group, to get what we want.

link


> 
> You can just use the CTA heading size, it automatically adjusts down to 52px on small screens. I've made that change in [47c579a](https://github.com/WordPress/wporg-parent-2021/commit/47c579ada39b5336ce919dee4311de51e1835849). Actually, as I'm writing this up, I realized we can just set that on the heading itself, so I've done that in [49c061a](https://github.com/WordPress/wporg-parent-2021/commit/49c061aa4c771a29a511f1a9f304abf0c9fc510b).
> 

Didn't realize this... this effect only happens in the published post. Nice finding...!

> I think this is good to merge now. There's something awkward happening at mid-sized screens (600 - 1080), but we can figure that out in a later PR.
> 
> Screen Shot 2022-08-04 at 6 31 15 PM

I can't reproduce this one, but I've opened an [issue](https://github.com/WordPress/wporg-parent-2021/issues/32) for this.

https://user-images.githubusercontent.com/18050944/182996578-f63dd7ed-5b00-4e98-8043-c60d27bb8439.mp4

> I think this is good to merge now.

Thanks for the help and the reply!

PR #22, opened 2022-08-02
Block Style: Add "Serif" style for paragraphs

### LINKED ISSUE #21: Paragraphs need EB Garamond font style option
Headings are EB Garamond, and body text is Inter. These are set in the `theme.json`. In a few cases, elements in the mockup use the opposite font though, so we need to be able to control that somehow.

This should be a `p`, but it's using the heading font.
Screen Shot 2022-07-28 at 10 32 40 AM 1

Maybe block styles? Or is there a block-level font family control we can use?

ryelle:
Fixes #21. Add a block style to switch the font of a paragraph block.

### Screenshots

Block style is available on the paragraph block:

Screen Shot 2022-08-02 at 12 50 00 PM

| Before | After |
|--------|-------|
|  |  |

1. Create or edit a post/page
2. Add a paragraph
3. Switch to the serif block style
4. The font should change in the editor
5. Save the post, view it
6. The font should be serif for that paragraph on the front end

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -169,3 +169,9 @@
 		line-height: var(--wporg--style--feature--li-height);
 	}
 }
+
+.is-style-serif {
+	font-family: var(--wp--custom--heading--typography--font-family);
+	font-size: var(--wp--preset--font-size--heading-5);
+	line-height: var(--wp--custom--heading--level-5--typography--line-height);
+}
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -85,6 +85,15 @@ function setup_block_styles() {
 			'label'        => __( 'Features', 'wporg' ),
 		)
 	);
+
+	register_block_style(
+		'core/paragraph',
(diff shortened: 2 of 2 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,9 @@
 		}
 	}
 }
+
+.is-style-serif {
+	font-family: var(--wp--custom--heading--typography--font-family);
+	font-size: var(--wp--preset--font-size--heading-5);
+	line-height: var(--wp--custom--heading--level-5--typography--line-height);
```
adamwoodnz, line 176: I realise these match our use case, but it feels strange to be setting them too; I'd expect to only be changing the `font-family`. Is there a reason not to use the standard typography text size controls for this?

adamwoodnz (commented):
This works for me, just one question :)

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: This just sets up the default, because I think it's safe to assume that when someone wants the serif size, they also want the slightly larger font. The typography controls will still work to set other sizes if this assumption is wrong.

adamwoodnz (approved):
LGTM :shipit:

PR #28, opened 2022-08-03
Block Style: Add Features list style

### LINKED ISSUE #32: Block style: Feature list with animation
This is a list of items (in the future these will be links). The text is oversized and positioned below the baseline, but on hover it animates up and darkens. See [the Releases page on News for an example](https://wordpress.org/news/category/releases/).

Note: the default color will be updated to pass AA color contrast.

Screen Shot 2022-07-28 at 6 47 36 PM

I think this can be [a block style](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-styles/) on the List block.

ryelle:
Fixes https://github.com/WordPress/wporg-main-2022/issues/32 — this adds a block style to the list block. Using this style bumps the font size, adds a bottom border, and offsets the content below the baseline slightly. On hover (or focus, if there's a link), the line animates slightly up to be fully visible.

### Screenshots

Editor

editor

Front end

default

With "Intuitive" hovered.

hover

Small screen

Screen Shot 2022-08-03 at 19 39 09

1. Add a list to a page, select "Features" in the sidebar.
2. Add items to the list
3. It should look like the screenshots

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -121,3 +121,50 @@
 		}
 	}
 }
+
+.is-style-features {
+	--wporg--style--feature--offset: 0.2em;
+	padding: 0;
+	list-style: none;
+	font-size: var(--wp--preset--font-size--heading-1);
+	font-weight: 200;
+	line-height: 1.1;
+
+	li {
+		overflow: hidden;
+		box-sizing: border-box;
+		height: 1.1em;
+		padding-top: var(--wporg--style--feature--offset);
+		padding-left: var(--wp--preset--spacing--50);
+		padding-right: var(--wp--preset--spacing--50);
+		color: #6682ff; // Hardcoded color for better contrast.
+		border-bottom: 1px solid var(--wp--preset--color--blueberry-3);
+		transition: all 0.2s ease-in-out;
+
+		a {
+			text-decoration: none;
+		}
+
+		&:hover,
+		&:focus,
+		&:focus-within {
+			padding-top: 0;
+			color: var(--wp--preset--color--blueberry-1);
+		}
+	}
+
+	// Make sure to remove padding from blocks with background colors.
+	&.has-background {
+		padding: 0;
+	}
+
+	// Use the set text color, if one exists.
+	&.has-text-color li {
+		color: inherit;
+	}
+
+	// Blocks with set font sizes should still use this line height.
+	&[class*="-font-size"] {
+		line-height: 1.1;
+	}
+}
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -131,7 +131,7 @@
 	font-weight: 200;
 	line-height: var(--wporg--style--feature--li-height);
 
-	li {
+	> li {
 		overflow: hidden;
 		box-sizing: border-box;
 		height: calc(var(--wporg--style--feature--li-height) * 1em);
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -124,16 +124,17 @@
 
 .is-style-features {
 	--wporg--style--feature--offset: 0.2em;
+	--wporg--style--feature--li-height: 1.1;
 	padding: 0;
 	list-style: none;
 	font-size: var(--wp--preset--font-size--heading-1);
 	font-weight: 200;
-	line-height: 1.1;
+	line-height: var(--wporg--style--feature--li-height);
 
 	li {
 		overflow: hidden;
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,50 @@
...
+		box-sizing: border-box;
+		height: 1.1em;
+		padding-top: var(--wporg--style--feature--offset);
+		padding-left: var(--wp--preset--spacing--50);
+		padding-right: var(--wp--preset--spacing--50);
+		color: #6682ff; // Hardcoded color for better contrast.
+		border-bottom: 1px solid var(--wp--preset--color--blueberry-3);
+		transition: all 0.2s ease-in-out;
+
+		a {
```
adamwoodnz: `li` and `a` tend to be quite common elements in pages. Worth using direct descendant selectors for a minor performance gain (CSS is parsed from right to left)?

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,50 @@
...
+			text-decoration: none;
```
adamwoodnz, line 146: Minor, but if I add a link this underline still shows in the editor on hover

Screen Shot 2022-08-04 at 4 34 55 PM

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,50 @@
...
+
+		&:hover,
+		&:focus,
+		&:focus-within {
+			padding-top: 0;
+			color: var(--wp--preset--color--blueberry-1);
+		}
+	}
+
+	// Make sure to remove padding from blocks with background colors.
```
adamwoodnz, line 157: Tried to test this but couldn't add a background color to an `li`. Or am I misunderstanding...

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -121,3 +121,50 @@
...
 	}
 }
+
+.is-style-features {
+	--wporg--style--feature--offset: 0.2em;
+	padding: 0;
+	list-style: none;
+	font-size: var(--wp--preset--font-size--heading-1);
+	font-weight: 200;
+	line-height: 1.1;
```
adamwoodnz: Variablize this? It's reused on line 168

adamwoodnz (approved):
Works for me in Chrome, FF and Safari

A few minor comments inline

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
adamwoodnz: This would leave the styles more heavily coupled to the markup structure though. We seem to have to attach a lot of styles to tags rather than classes, which are more robust and performant.

ryelle: I think we can be confident that the list block markup won't change, since it's a simple `ol,ul > li`, so I see no harm in tying to the structure. Also, there are no classes applied to the list items, so the only way to select them is through the tags. I can add a descendant selector to the `li`, though. The `a` i think I'll leave as-is, in case someone's bolded and linked the text or something :)

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: You can add a background to the list block itself in the sidebar

Screen Shot 2022-08-04 at 11 20 28 AM

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: Oh yep, it's not as visible for me in FF. Thanks for catching that.
Screen Shot 2022-08-04 at 11 20 11 AM

ryelle: Looks like this is fixed with the link fix that came out in GB 13.8, once I ran `composer update` to get updated, no more underline:

Screen Shot 2022-08-04 at 11 40 04 AM

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: Good idea - it's actually used 3 times, because the `height: 1.1em` is meant to be the same value too. Done in https://github.com/WordPress/wporg-parent-2021/pull/28/commits/479abdc78f45062cab882b8c66f6a1e74f37fce6

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: Done in https://github.com/WordPress/wporg-parent-2021/pull/28/commits/27b7678d73ed3bd7e57199e4c8c78dc8258aaa80

PR #31, opened 2022-08-05
Fix heading-with-arrow Heading incorrect font-size

outdoor2kode:
It seems it still needs adding a class name to `<h2>` to make font-size work correctly.


ryelle (approved):
Good catch, I missed this in my update 👍🏻

PR #33, opened 2022-08-05
Block Style: Add List of links list style

outdoor2kode:
Fixes https://github.com/WordPress/wporg-main-2022/issues/29.

### Screencasts

https://user-images.githubusercontent.com/18050944/183087412-755b7667-3f94-474e-ae10-3aef8110aade.mov

### How to test the changes in this Pull Request:

1.Add a list to a page, select "List of lists" in the sidebar.
2. Add items to the list.
3. It should look like the screencasts

PR #34, opened 2022-08-05
Block Style: Links list style (alternative)

### LINKED ISSUE #29: Block Style: List of links
This should be possible with [a block style](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-styles/) to add a class to the List block, and the content in the list block is just a list of links.

| Mockup | Current staging site |
|----------|---------|
| Screen Shot 2022-07-28 at 12 58 02 PM | Screen Shot 2022-07-28 at 1 03 17 PM |

ryelle:
Fixes https://github.com/WordPress/wporg-main-2022/issues/29

This is an alternative to #33 — I wanted to see how we could support custom colors and what should happen if this block is inside colorful groups. The block itself, with no settings, just uses the default text & link colors.

Screen Shot 2022-08-05 at 12 26 32 PM

When it's in a container with white text, the border goes to 20% opacity to match the design. This is probably the context it will appear on the site in.

Screen Shot 2022-08-05 at 12 40 07 PM

Otherwise it matches the current text color. The border is full-opacity here because we can't turn down the opacity without knowing what color it is. White is the special case to match the design.

Screen Shot 2022-08-05 at 12 40 51 PM

### Screenshots

| Editor | Frontend |
|--------|-------|
|  |  |

1. Add a list block
2. Set text colors & background colors, it should update to match
3. Reset the colors
4. Put it in a group or columns block
5. Set some colors on the container block, it should update to match

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -170,6 +170,47 @@
 	}
 }
 
+.is-style-links-list {
+	padding: 0;
+	list-style: none;
+	font-size: var(--wp--preset--font-size--extra-large);
+	line-height: 1.125;
+
+	> li {
+		padding: 14px 0;
+		// Use the current text color, to support different text/background color combos.
+		border-bottom: 1px solid currentColor;
+
+		a {
+			text-decoration: none;
+
+			&:hover,
+			&:focus {
+				text-decoration-line: underline;
+				text-decoration-thickness: 1px;
+				text-underline-offset: 0.15em;
+			}
+		}
+
+		&:first-of-type {
+			border-top: 1px solid currentColor;
+		}
+	}
+
+	// Remove padding from blocks with background colors.
+	&.has-background {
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -170,6 +170,47 @@
...
+	// Remove padding from blocks with background colors.
+	&.has-background {
+		padding: 0;
+	}
+
+	// If the text color is set to white, the border should be semi-transparent.
+	// This matches the design, when the list is used inside a blue background.
+	&.has-white-color > li,
+	.has-white-color > & > li,
+	.has-white-color *:not(.has-text-color) & > li {
```
outdoor2kode, line 209: 👍 👍 👍 
I was trying to set the border to semi-transparent as well when `has-link-color` exists, but it seems it's set without also specifying color..

outdoor2kode (approved):
Tested and things worked well on my side. Looks good to me! I'll close #33.

PR #36, opened 2022-08-05
Block Style: Add a "With Arrow" style to the Heading block

### LINKED ISSUE #32: Styles go off at mid-sized screens for heading-and-arrow pattern

### LINKED ISSUE #35: Pattern: Heading with Arrow has background on hover
This was the style for News, but I don't think it's expected here. Remove the background on hover, keep the animation and underline.

Screen Shot 2022-08-05 at 2 44 39 PM

ryelle:
Fixes #32, fixes #35. This tries to add a block style for the "with arrow" heading, so we can add it in the editor without needing a special CSS class. I've also tried creating it with just CSS, so that it can reflect the heading text color, in case it's needed in other color schemes.

This fixes the alignment on mid-sized screens by dropping down to the 2nd line at 960px. This seemed to be a natural place to drop off when the text is so large.

The hover color noted in #35 is gone here — the underline for links should be enough visual feedback.

### Screenshots

| Editor | Frontend | Small Screen |
|--------|-------|-------|
| Screen Shot 2022-08-05 at 16 38 13 | Screen Shot 2022-08-05 at 16 38 16 | Screen Shot 2022-08-05 at 16 39 07 |

https://user-images.githubusercontent.com/541093/183195283-f29c0ac3-6884-4d80-9a5c-7308b43ad768.mp4

https://user-images.githubusercontent.com/541093/183195302-85e4c36e-9456-47c3-85e8-dc44b1091fb1.mp4

1. Add a heading block to a page
2. Switch to the With Arrow style
3. It should render the arrow in the editor
4. Save it, and view the frontend, you should see the same on the frontend
5. Back in the editor, try different heading levels, adjust the settings in the block UI, it should all work

Test with the pattern too:

1. Add the "Heading with arrow" pattern
2. It should add a blue box with white text
3. Edit the content in the pattern, it should behave as expected

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
@@ -6,12 +6,12 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","backgroundColor":"blueberry-1"} -->
-<div class="wp-block-group alignfull has-blueberry-1-background-color has-background">
-	<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"97px","bottom":"97px"},"blockGap":"0px"}},"className":"wporg-pattern__heading-with-arrow"} -->
-	<div class="wp-block-group alignfull wporg-pattern__heading-with-arrow has-link-color" style="padding-top:97px;padding-bottom:97px">
-		<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"fontSize":"heading-cta","fontFamily":"inter"} -->
-		<h2 class="has-inter-font-family has-heading-cta-font-size" style="font-style:normal;font-weight:300;letter-spacing:-0.02em;line-height:1.1"><a href="https://wordpress.org/download/">Get started</a></h2>
+<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|60"}}},"backgroundColor":"blueberry-1","textColor":"white","layout":{"inherit":true}} -->
+<div class="wp-block-group alignfull has-white-color has-blueberry-1-background-color has-text-color has-background has-link-color" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)">
+	<!-- wp:group {"align":"wide"} -->
+	<div class="wp-block-group alignwide">
+		<!-- wp:heading {"align":"wide","className":"is-style-with-arrow","fontSize":"heading-1","fontFamily":"inter"} -->
+		<h2 class="alignwide is-style-with-arrow has-inter-font-family has-heading-1-font-size"><a href="https://wordpress.org/download/">Get started</a></h2>
 		<!-- /wp:heading -->
 	</div>
 	<!-- /wp:group -->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
@@ -10,8 +10,8 @@
 <div class="wp-block-group alignfull has-white-color has-blueberry-1-background-color has-text-color has-background has-link-color" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)">
 	<!-- wp:group {"align":"wide"} -->
 	<div class="wp-block-group alignwide">
-		<!-- wp:heading {"align":"wide","className":"is-style-with-arrow","fontSize":"heading-1","fontFamily":"inter"} -->
-		<h2 class="alignwide is-style-with-arrow has-inter-font-family has-heading-1-font-size"><a href="https://wordpress.org/download/">Get started</a></h2>
+		<!-- wp:heading {"align":"wide","className":"is-style-with-arrow","fontSize":"heading-cta","fontFamily":"inter"} -->
+		<h2 class="alignwide is-style-with-arrow has-inter-font-family has-heading-cta-font-size"><a href="https://wordpress.org/download/">Get started</a></h2>
 		<!-- /wp:heading -->
 	</div>
 	<!-- /wp:group -->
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
```diff
@@ -6,12 +6,12 @@
...
+<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|60"}}},"backgroundColor":"blueberry-1","textColor":"white","layout":{"inherit":true}} -->
+<div class="wp-block-group alignfull has-white-color has-blueberry-1-background-color has-text-color has-background has-link-color" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)">
+	<!-- wp:group {"align":"wide"} -->
+	<div class="wp-block-group alignwide">
+		<!-- wp:heading {"align":"wide","className":"is-style-with-arrow","fontSize":"heading-1","fontFamily":"inter"} -->
+		<h2 class="alignwide is-style-with-arrow has-inter-font-family has-heading-1-font-size"><a href="https://wordpress.org/download/">Get started</a></h2>
```
ryelle: This should be the CTA font size

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
```diff
@@ -6,12 +6,12 @@
...
-		<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"300","lineHeight":"1.1","letterSpacing":"-0.02em"}},"fontSize":"heading-cta","fontFamily":"inter"} -->
-		<h2 class="has-inter-font-family has-heading-cta-font-size" style="font-style:normal;font-weight:300;letter-spacing:-0.02em;line-height:1.1"><a href="https://wordpress.org/download/">Get started</a></h2>
+<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|60"}}},"backgroundColor":"blueberry-1","textColor":"white","layout":{"inherit":true}} -->
+<div class="wp-block-group alignfull has-white-color has-blueberry-1-background-color has-text-color has-background has-link-color" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)">
+	<!-- wp:group {"align":"wide"} -->
```
outdoor2kode, line 11: I think we can remove this one as it has no effect anymore. Originally I set two group containers for hover color, but we no longer need it.

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
outdoor2kode: But why do we use `has-heading-1-font-size` here?

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -176,72 +177,108 @@
...
 }
 
-.wporg-pattern__heading-with-arrow {
+.is-style-with-arrow {
+	// Note, these are Sass variables to make the math easier. While we could do
+	// this with `calc()`s, it would be even more unruly.
+	$arrow_size: 1em;
+	// Once rotated, the diagonal of the square is the hypotenuse of an isosceles
+	// triangle, which can be calculated by multiplying the side × √2. We only need
+	// the halfway point, so divide that by 2.
```
outdoor2kode, line 231: 👍

outdoor2kode (commented):
https://user-images.githubusercontent.com/18050944/183478681-da3ae06a-19a7-453a-a9c8-b59baee4426f.mp4

For a bigger screen, I guess since we're using `vw` in `min-width`, it would squeeze the link text space. It seems we don't need `vw` or we can use it in `a[href]` instead?

Other than that, I think it looks good to me! Thanks for the fix and adding it to block style, it looks better.

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
ryelle: It was a mistake I noticed after I made the PR, and haven't circled back yet to fix it

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
outdoor2kode: I just realized that since we already have `padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)` in parent group, so there's still padding existing after removing this group.

outdoor2kode: Do you think changing from using a variable to an exact number is better?

`"spacing":{"padding":{"top":"80px","right":"80px","bottom":"80px","left":"80px"}}},`
`style="padding-top:80px;padding-right:80px;padding-bottom:80px;padding-left:80px"`

Upside: Users can see padding in GB UI tools.
image
Downside: We can't configure it through theme.json directly.

ryelle: There are two group containers here so that the top-level one can be full-width across the screen, and the child group is `wide` to give the heading a max-width. This also lets authors put content inside the wide group so it lines up correctly with the heading, like this example:

Screen Shot 2022-08-08 at 3 27 53 PM
 
> Do you think changing from using a variable to an exact number is better?

No, because `--wp--preset--spacing--60` is not an exact number, it's [a responsive value using clamp](https://github.com/WordPress/wporg-parent-2021/blob/fb178ec3bce9ef150ed13701ec2db439e6ea6035/source/wp-content/themes/wporg-parent-2021/theme.json#L487-L491). Users can still override this by setting values, but this way the default value adjusts to the screen size. Eventually this method should be supported by the UI too (there's an in-progress PR here: https://github.com/WordPress/gutenberg/pull/42173), so I'd prefer to stick with this.

ryelle:
I also adjusted the CSS a little to use grid for all screen sizes, so that I could use the `fr` unit to span the rest of the content area. This way the arrow is always the full length, even if the text is very short.

Screen Shot 2022-08-08 at 3 52 08 PM

source/wp-content/themes/wporg-parent-2021/patterns/heading-with-arrow.php
outdoor2kode: > This also lets authors put content inside the wide group so it lines up correctly with the heading
I see, thanks for the clarification!

>  there's an in-progress PR here: https://github.com/WordPress/gutenberg/pull/42173
👍!

ryelle:
I'm going to merge this now for the design review deadline, we can follow up with any changes in other PRs.

PR #39, opened 2022-08-17
Block Style: Add a "Dots" navigation block style

### LINKED ISSUE #94: Add missing links to the download page
We forgot linking to several important pages from the download page:

- https://wordpress.org/download/releases
- https://wordpress.org/download/beta-nightly
- https://wordpress.org/download/counter
- https://wordpress.org/download/source

We've updated [the designs](https://www.figma.com/file/WHhRwVc4swYnPKdyVWj3du/Home-%26-Download-Pages?node-id=970%3A9767) to add them under the download links

image

ryelle:
Fixes https://github.com/WordPress/wporg-main-2022/issues/94 — this updates the default navigation styles, and adds a new block style called "Dots." The updates to the default styles just ensures that the navigation uses the right underline style. The Dots style adds separator dots between top-level menu items.

Edit: I removed the default color, because it conflicted with the global header link colors. I didn't re-take the screenshots, so when testing, these links will be black (unless you set a text color).

Edit 2: The dots style now also removes the block gap, since the dots function as block spacers.

### Screenshots

| Editor | Frontend |
|--------|-------|
|  |  |

1. Add a navigation menu to a page
2. The links should not be underlined by default, and underline on hover
3. Try the dots style
4. There should now be grey dots separating the links

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -328,16 +328,20 @@
 	}
 }
 
-.wp-block-navigation.is-style-dots > .wp-block-navigation__container > .wp-block-navigation-item {
-	&::after {
-		content: "·";
-		color: var(--wp--preset--color--charcoal-4);
-		width: 1em;
-		text-align: center;
-		position: relative;
-	}
+.wp-block-navigation.is-style-dots {
+	.wp-block-navigation-item {
+		&::after {
+			content: "·";
+			color: var(--wp--preset--color--charcoal-4);
+			width: 1em;
+			text-align: center;
+			position: relative;
+		}
 
-	&:last-of-type::after {
-		content: "";
+		&.wp-block-navigation-submenu::after,
+		.wp-block-navigation-item::after,
+		&:last-of-type::after {
+			content: "";
+		}
 	}
 }
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -327,3 +327,17 @@
 		}
 	}
 }
+
+.wp-block-navigation.is-style-dots .wp-block-navigation-item {
+	&::after {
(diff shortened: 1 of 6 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -327,3 +327,17 @@
 		}
 	}
 }
+
+.wp-block-navigation.is-style-dots > .wp-block-navigation__container > .wp-block-navigation-item {
```
adamwoodnz: This doesn't completely work for me for some reason. I see the option in the sidebar and it displays as long as I set the overlay menu to false. The other options result in a `wp-block-navigation__responsive-container` and no `wp-block-navigation__container` so the style doesn't apply.

Note that it shows the last dot while the block is selected.

Screen Shot 2022-08-18 at 11 29 38 AM

In the frontend no `is-style-dots` class is applied to page level navigation, but it does work in the post content. I've tried toggling all the settings mentioned above but nothing changes. Can't work it out. Looking in the database didn't help me unfortunately.

Screen Shot 2022-08-18 at 12 02 02 PM

adamwoodnz (commented):
LGTM in post content but not at page level, see comment inline

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: > This doesn't completely work for me for some reason.

Ah, I only tested non-responsive menus, the markdown changes a lot on the responsive ones. It should work correctly now.

> it shows the last dot while the block is selected.

Yeah, the appender becomes the `last-child`, so the CSS that would remove it doesn't apply. IMO this is pretty minor, maybe we can fix it in a following PR?

> In the frontend no is-style-dots class is applied to page level navigation, but it does work in the post content. 

This is working for me now, but it seemed to be fixed when I fixed the responsive nav issue, so I don't know if that was the same problem for you.

adamwoodnz (approved):
:shipit: LGTM

PR #41, opened 2022-08-22
Make CTA heading arrow clickable

### LINKED ISSUE #40: Heading with Arrow: Arrow area should be a clickable part of link
Initially reported here: https://meta.trac.wordpress.org/ticket/6447

When this pattern is used, only the text is clickable despite the arrow appearing to be part of the link.

680FC577-373E-4CE4-9ED6-DF426D98700B

Maybe we _should_ assume that a heading using this pattern will always have a link, and add the arrow on `a`.

adamwoodnz:
Fixes #40 by applying the arrow style to the anchor within the heading tag, enabling it to be clicked. As mentioned in the ticket this assumes that this style of heading will always be a call to action and hence have a link applied.

### Screenshots

| Editor | FE Desktop | FE Small
|--------|--------|--------|
| Screen Shot 2022-08-22 at 1 53 44 PM | Screen Shot 2022-08-22 at 1 54 00 PM | Screen Shot 2022-08-22 at 1 54 18 PM |

1. Add a heading to a post or page
2. Select the CTA option
3. Observe that no arrow appears
4. Add a link to the heading text
5. Observe that the arrow style is applied

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -117,7 +117,7 @@ function setup_block_styles() {
 		'core/heading',
 		array(
 			'name'         => 'with-arrow',
-			'label'        => __( 'CTA', 'wporg' ),
+			'label'        => __( 'Link arrow', 'wporg' ),
 			'style_handle' => STYLE_HANDLE,
 		)
 	);
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -117,7 +117,7 @@ function setup_block_styles() {
 		'core/heading',
 		array(
 			'name'         => 'with-arrow',
-			'label'        => __( 'Link arrow', 'wporg' ),
+			'label'        => __( 'Link & Arrow', 'wporg' ),
 			'style_handle' => STYLE_HANDLE,
 		)
 	);
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -117,7 +117,7 @@ function setup_block_styles() {
 		'core/heading',
 		array(
 			'name'         => 'with-arrow',
-			'label'        => __( 'With Arrow', 'wporg' ),
+			'label'        => __( 'CTA', 'wporg' ),
 			'style_handle' => STYLE_HANDLE,
 		)
 	);
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -248,13 +248,6 @@
 		gap: 0.5em;
 	}
 
-	&:hover,
-	&:focus {
-		text-decoration-line: underline;
-		text-decoration-thickness: 1px;
-		text-underline-offset: 0.15em;
-	}
-
 	&::before,
 	&::after {
 		content: "";
@@ -314,7 +307,12 @@
 	}
 
 	&:hover,
+	&:focus,
 	&:focus-within {
+		text-decoration-line: underline;
+		text-decoration-thickness: 1px;
+		text-underline-offset: 0.15em;
+		
 		&::before {
 			transform: translateX(#{math.div($arrow_size, 2)}) rotate(45deg);
 		}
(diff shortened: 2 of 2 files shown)


source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
```diff
@@ -117,7 +117,7 @@ function setup_block_styles() {
 		'core/heading',
 		array(
 			'name'         => 'with-arrow',
-			'label'        => __( 'With Arrow', 'wporg' ),
+			'label'        => __( 'CTA', 'wporg' ),
```
adamwoodnz: This is up for debate but I think it makes it a little more obvious that this heading style is expected to have a link applied

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -222,7 +222,7 @@
 	line-height: var(--wp--custom--body--short-text--typography--line-height);
 }
 
-.is-style-with-arrow {
+.is-style-with-arrow a[href]:first-of-type {
```
adamwoodnz, line 225: The `first-of-type` selector prevents multiple arrows with multiple anchors. The layout behaviour can be a little unexpected if the anchor is used on a sub string of the heading, but I'm not sure this is worth trying to mitigate:

Screen Shot 2022-08-22 at 2 31 48 PM

source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
ryelle: I'd prefer to keep it descriptive of the pattern, since we might have other styles that are "calls to action". How about "Link with arrow" instead? Then the class name is still consistent.

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -304,18 +313,6 @@
...
-		&:hover,
-		&:focus {
-			text-decoration-line: underline;
-			text-decoration-thickness: 1px;
-			text-underline-offset: 0.15em;
-		}
-	}
-
 	&:hover,
 	&:focus-within {
```
ryelle: You can merge the hover & focus styles with the styles added at line 251

source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
ryelle: I agree, we don't need to address that issue now. If it becomes something site editors need to do, we can figure it out then.

ryelle (commented):
The heading now has some extra space under it in the editor view - it's not a new line or extra content, it seems to be caused by the link somehow. (I changed the colors so the outline was easier to see, it happens on the default pattern too)

Screen Shot 2022-08-22 at 1 04 39 PM

I don't see this problem in Chrome, so maybe it's just a Firefox issue. If it's not simple to fix, we can follow up with a separate issue/PR so that the frontend link fix can be deployed.

adamwoodnz:
> The heading now has some extra space under it in the editor view - it's not a new line or extra content, it seems to be caused by the link somehow.

Yep something to do with `display: grid`, still investigating....

source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
adamwoodnz: I did try that but the text is too long and becomes truncated

Screen Shot 2022-08-23 at 4 40 15 PM

'Link arrow' fits

Screen Shot 2022-08-23 at 4 42 32 PM

Or I guess we could add styles to wrap the text?

adamwoodnz:
> > The heading now has some extra space under it in the editor view - it's not a new line or extra content, it seems to be caused by the link somehow.
> 
> Yep something to do with `display: grid`, still investigating....

@ryelle I think the right place to fix this is via [editor.scss](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/sass/editor.scss) so that we're not doing unnecessary things in the frontend, and as mentioned elsewhere I can't get those styles to work. Even in the inline styles I don't see any of the rules from that file.

Maybe as you say we should land this as is and work on a follow up PR once I'm able to fix via editor styles.

source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
ryelle: No, if we override the styles here it might have other effects/conflicts in the future. I tried Link & Arrow and that isn't cut off 👍🏻

ryelle (approved):
Looks good now 👍🏻 

> Maybe as you say we should land this as is and work on a follow up PR once I'm able to fix via editor styles.

Yeah, let's follow up on this later.

PR #42, opened 2022-08-23
Make style of 'Link & arrow' heading in Firefox editor match frontend

### LINKED ISSUE #40: Heading with Arrow: Arrow area should be a clickable part of link

adamwoodnz:
Fixes #40 by overriding inline styles added by editor causing extra space under heading blocks with the option 'Link & arrow'.

Props @ryelle 

#### Frontend (note no space under heading and no single space before CTA on the bottom multiline version)

Screen Shot 2022-08-24 at 10 49 15 AM

#### Editor

| Before | After |
|--------|-------|
| Screen Shot 2022-08-24 at 10 47 48 AM | Screen Shot 2022-08-24 at 10 48 54 AM |

1. Open the editor in Firefox
2. Create a post and add a heading block
3. Below the heading add any other block
4. Choose the 'Link & arrow' option
5. There shouldn't be excessive space below the heading
6. Check that the display in the editor matches the frontend
7. Check other supported browsers

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_heading.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_heading.scss
@@ -1,0 +1,5 @@
+.wp-block-heading {
+	&.is-style-with-arrow {
+		white-space: revert !important;
+	}
+}
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_heading.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_heading.scss
@@ -1,5 +1,0 @@
-.wp-block-heading {
-	&.is-style-with-arrow {
-		white-space: revert !important;
-	}
-}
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-parent-2021/sass/blocks/_heading.scss
```diff
@@ -0,0 +1,5 @@
+.wp-block-heading {
```
ryelle: Separate files like this are intended to be added to the frontend too, so I've moved this code directly into `editor.scss`. The other files imported into `editor.scss`, `button` and `navigation`, are frontend styles, added to ensure the styling is correct across both.

ryelle (approved):
Thanks for fixing this, it's working properly in FF now 👍🏻

PR #48, opened 2022-10-12
Block styles: Add new "Brush Stroke" style for Group blocks

ryelle:
See https://github.com/WordPress/wporg-main-2022/issues/129, https://github.com/WordPress/wporg-showcase/issues/1 — This adds a new block style on the Group block called "Brush Stroke." When selected, it adds a mask to the group to cut out a brush stroke from the background color. A background color is needed to be visible, but any background should work.

Note: this is just for the background, the nav menu dropdown will be done separately (maybe in wporg-main-2022? not totally sure how it'll work yet).

### Screenshots

With the main child theme

| Desktop | Mobile |
|--------|-------|
| brush-banner | brush-banner-small |

editor

With other colors
multiple

1. Build the branch
2. Try using it in any post content

To test on the main child theme

1. Switch to the branch in https://github.com/WordPress/wporg-main-2022/pull/131
2. Swap out the first group in any of the download subpage patterns with the following code
3. View the pages, the blue bar should appear with the brush stroke
4. The bar should also dock to the top when scrolling, like on News (when >890px wide)

```html
<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"var:custom|brush-stroke|spacing|height","right":"var:preset|spacing|20","bottom":"0","left":"var:preset|spacing|20"}}},"backgroundColor":"pomegrade-1","textColor":"white","className":"is-style-brush-stroke","layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull is-style-brush-stroke has-black-color has-pomegrade-1-background-color has-text-color has-background has-link-color" style="padding-top:var(--wp--custom--brush-stroke--spacing--height);padding-right:var(--wp--preset--spacing--20);padding-bottom:0;padding-left:var(--wp--preset--spacing--20)"><!-- wp:group {"align":"wide","layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
<div class="wp-block-group alignwide"><!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"700","lineHeight":1.6}},"fontSize":"normal"} -->
<p class="has-normal-font-size" style="font-style:normal;font-weight:700;line-height:1.6">Download</p>
<!-- /wp:paragraph -->

<!-- wp:navigation {"ref":11576,"overlayMenu":"never","style":{"typography":{"lineHeight":1.6}}} /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -355,3 +355,36 @@
...
+		mask-repeat: no-repeat;
+		mask-size: cover;
+		mask-position: bottom;
+		background-color: inherit;
+	}
+}
+
+@media (min-width: 890px) {
+	// Only add this for pages with the block on it.
+	html:has(main > .is-style-brush-stroke) {
```
ryelle, line 380: caniuse puts [:has support](https://caniuse.com/?search=has) pretty low, compared to what we usually try to support (no Firefox support at all 😞), but I think this can be considered a progressive enhancement. 

Open to other suggestions though.

adamwoodnz (approved):
Nice work, LGTM :shipit: 

Works for me in all scenarios mentioned above.

PR #50, opened 2022-10-25
Add post-navigation pattern

StevenDufresne:
This PR adds a pattern for navigating to previous/next posts. I am adding this pattern here because it now shows up in 2 themes, `wporg-showcase-2022` ([Showcase example](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=656%3A12482)), and `wporg-news-2021` ([News Site Example](https://wordpress.org/news/2022/10/wordpress-6-0-3-security-release/)).

See https://github.com/WordPress/wporg-showcase-2022/issues/6

### Screenshots

Disregard the page layout, it's an incomplete part of WordPress/wporg-showcase-2022.

| Desktop |  Table | Mobile |
|--------|-------|-------| 
|  Screen Shot 2022-10-25 at 1 58 17 PM | Screen Shot 2022-10-25 at 1 58 23 PM | Screen Shot 2022-10-25 at 1 58 37 PM

1. Ensure you have multiple posts.
2. Add `<!-- wp:pattern {"slug":"wporg-parent-2021/post-navigation"} /-->` to a single post view

# Notes
- It includes inline CSS styles which were copied directly out of the block editor. I'm not sure if that's the best approach. 🤔 
- In wporg-main-2021, the wrapper container is a `nav` tag. Since this is using default block editor styles, the `<Row>` component doesn't allow for that selection. I could wrap the group with a nav for semantics sake. Alternatively i can stop relying on the default block editor built styles. **Thoughts**?

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
@@ -8,17 +8,7 @@
 ?>
 
 <!-- wp:group {"style":{"border":{"top":{"color":"var:preset|color|light-grey-1","width":"1px"}},"spacing":{"padding":{"top":"var:preset|spacing|20","right":"var:preset|spacing|20","bottom":"var:preset|spacing|20","left":"var:preset|spacing|20"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
-<div
-	class="wp-block-group"
-	style="
-		border-top-color: var( --wp--preset--color--light-grey-1 );
-		border-top-width: 1px;
-		padding-top: var( --wp--preset--spacing--20 );
-		padding-right: var( --wp--preset--spacing--20 );
-		padding-bottom: var( --wp--preset--spacing--20 );
-		padding-left: var( --wp--preset--spacing--20 );
-	"
->
+<div class="wp-block-group" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-width:1px;padding-top:var(--wp--preset--spacing--20);padding-right:var(--wp--preset--spacing--20);padding-bottom:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--20)">
 	<!-- wp:post-navigation-link {"type":"previous","showTitle":true,"label":"<?php esc_attr_e( 'Previous', 'wporg' ); ?>","linkLabel":true} /-->
 	<!-- wp:post-navigation-link {"type":"next","showTitle":true,"label":"<?php esc_attr_e( 'Next', 'wporg' ); ?>","linkLabel":true} /-->
 </div>
--- a/source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Title: Post Navigation
+ * Slug: wporg-parent-2021/post-navigation
+ * Inserter: no
+ */
+
+?>
+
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Title: Post Navigation
+ * Slug: wporg-parent-2021/post-navigation
+ * Inserter: no
```
ryelle:
```suggestion
 * Category: wporg
```

Since this isn't a content-syncing pattern, it could be used in the editor (even just by Neso building pages). Removing `Inserter: no` doesn't stop you from still using it in `<!-- wp:pattern` code.

source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
```diff
@@ -0,0 +1,25 @@
...
+	class="wp-block-group"
+	style="
+		border-top-color: var( --wp--preset--color--light-grey-1 );
+		border-top-width: 1px;
+		padding-top: var( --wp--preset--spacing--20 );
+		padding-right: var( --wp--preset--spacing--20 );
+		padding-bottom: var( --wp--preset--spacing--20 );
+		padding-left: var( --wp--preset--spacing--20 );
+	"
+>
```
ryelle: This auto-formatting unfortunately breaks the block validation when loaded in the editor, and while it doesn't have the same issue in `wp:pattern` blocks it's better to be safe.

```suggestion
<div class="wp-block-group" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-width:1px;padding-top:var(--wp--preset--spacing--20);padding-right:var(--wp--preset--spacing--20);padding-bottom:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--20)">
```

I'm also guessing this is the inline CSS you mentioned copying? This is fine, it's part of the block itself. Removing it would also cause validation errors (and wouldn't load on the frontend — since this is a static block, the content here is what's used, the attributes in the JSON comment are only used by the editor).

source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
```diff
@@ -0,0 +1,25 @@
...
+	<!-- wp:post-navigation-link {"type":"previous","showTitle":true,"label":"<?php _e( 'Previous', 'wporg' ); ?>","linkLabel":true} /-->
+	<!-- wp:post-navigation-link {"type":"next","showTitle":true,"label":"<?php _e( 'Next', 'wporg' ); ?>","linkLabel":true} /-->
```
ryelle: Looks like we never set up a linter CI action here, but these are flagged by phpcs (`composer run lint source/wp-content/themes/wporg-parent-2021`).

```suggestion
	<!-- wp:post-navigation-link {"type":"previous","showTitle":true,"label":"<?php esc_attr_e( 'Previous', 'wporg' ); ?>","linkLabel":true} /-->
	<!-- wp:post-navigation-link {"type":"next","showTitle":true,"label":"<?php esc_attr_e( 'Next', 'wporg' ); ?>","linkLabel":true} /-->
```

ryelle (commented):
> In wporg-main-2021, the wrapper container is a nav tag.

Did you mean this? https://github.com/WordPress/wporg-news-2021/blob/15fa14f61cbaefc9fd83aa2914ff5a7d6c9b8cb8/source/wp-content/themes/wporg-news-2021/block-template-parts/post-navigation.html#L5-L8

That isn't a block, so it would definitely trigger an error. I don't think the `nav` element is as necessary in this kind of pagination (vs the 1 2 3… type), so I think it's fine as it is here.

source/wp-content/themes/wporg-parent-2021/patterns/post-navigation.php
StevenDufresne: Got it. Makes sense.

StevenDufresne:
> https://github.com/WordPress/wporg-news-2021/blob/15fa14f61cbaefc9fd83aa2914ff5a7d6c9b8cb8/source/wp-content/themes/wporg-news-2021/block-template-parts/post-navigation.html#L5-L8

Yes, too many themes in my head right now 😵 .

PR #53, opened 2022-11-15
Add an external link icon and css style.

StevenDufresne:
External links should display an external link icon. This PR adds an `svg`, taken from Gutenberg, and a class to target external links.

I wasn't exactly sure what color to give the icon. I went with a moderately dark grey color from our theme. 

See https://github.com/WordPress/wporg-showcase-2022/issues/42.

### Drawback of this approach
 Since we are only using CSS and not a block or something similar, users cannot see the external icon when viewing the post in the block editor. However, I don't expect there to be many external links, and doing more work should probably be considered only if we receive feedback from teams (maybe the documentation team may find that frustrating for example). 

### Screenshots

| Caption | Screenshot | 
| --- | --- |
|  Default | Screen Shot 2022-11-15 at 9 55 59 AM | 
|  With Element font-size: 40px | Screen Shot 2022-11-15 at 9 55 47 AM | 
|  Inline | Screen Shot 2022-11-15 at 10 45 32 AM | 
|  Alternative color option -- Not Implemented here | Screen Shot 2022-11-15 at 9 56 41 AM | 

- Add a link to some post content, add `external-link` class to the element. Expect to the the icon.
- Alternatively, you can add the class to the direct parent of the `<a>` 
  - This is done in case we don't have direct access to the `<a>` element.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -29,7 +29,10 @@ mark.has-inline-color {
 }
 
 // For links that navigate users to different site
-a:not([href*="wordpress.org"])::after {
+a.external-link::after,
+a.external-link-light::after,
+.external-link > a:not([href*="wordpress.org"])::after,
+.external-link-light > a:not([href*="wordpress.org"])::after {
 	content: "";
 	display: inline-flex;
 	background-color: var(--wp--preset--color--charcoal-4);
@@ -39,8 +42,12 @@ a:not([href*="wordpress.org"])::after {
 	vertical-align: middle;
 	-webkit-mask-image: url(../images/external-link-icon.svg);
 	mask-image: url(../images/external-link-icon.svg);
-	-webkit-mask-size: cover;
 	mask-size: cover;
+}
+
+a.external-link-light::after,
+.external-link-light > a:not([href*="wordpress.org"])::after {
+	background-color: var(--wp--preset--color--light-grey-2);
 }
 
 // Inline code tags should use the monospace font.
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -28,6 +28,22 @@ mark.has-inline-color {
 	}
 }
 
+// For links that navigate users to different site
+.external-link > a::after, // In case we don't access to <a> tag
+a.external-link::after {
+	content: "";
+	display: inline-flex;
+	background-color: var(--wp--preset--color--charcoal-4);
+	width: 1em;
+	height: 1em;
+	margin-left: 0.1em;
+	vertical-align: middle;
+	-webkit-mask-image: url(../images/external-link-icon.svg);
+	mask-image: url(../images/external-link-icon.svg);
+	-webkit-mask-size: cover;
+	mask-size: cover;
+}
+
 // Inline code tags should use the monospace font.
 code {
 	font-family: var(--wp--preset--font-family--monospace);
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
```diff
@@ -28,6 +28,22 @@ mark.has-inline-color {
 	}
 }
 
+// For links that navigate users to different site
+.external-link > a::after, // In case we don't access to <a> tag
+a.external-link::after {
```
ryelle: Instead of a custom class, could you use attribute selection to check if the `href` does not contain `wordpress.org`? I think this would work `a:not([href*="wordpress.org"])` ?

StevenDufresne: Yep, this appears to work on showcase (and a few other tests).

StevenDufresne: It's important to note that this is aggressive :)

**Footer**
Screen Shot 2022-11-17 at 12 35 33 PM

I'll need to run a more in-depth test before merging. Switching this back to a draft until I can.

StevenDufresne: After testing this patch using `wporg-main`, I think I'm tempted to move back to having an `external` class. I think we'll end up with a lot of places that have external URLs that we don't want to have the icon. 

Screen Shot 2022-11-22 at 3 59 11 PM
Screen Shot 2022-11-22 at 4 16 38 PM

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
```diff
@@ -28,6 +28,21 @@ mark.has-inline-color {
...
+	content: "";
+	display: inline-flex;
+	background-color: var(--wp--preset--color--charcoal-4);
+	width: 1em;
+	height: 1em;
+	margin-left: 0.1em;
+	vertical-align: middle;
+	-webkit-mask-image: url(../images/external-link-icon.svg);
+	mask-image: url(../images/external-link-icon.svg);
+	-webkit-mask-size: cover;
```
ryelle: The build process here has autoprefixer, so that should handle adding in the `-webkit-` properties.

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
ryelle: Fair enough… unfortunately there's no way to add a class to a link in the editor (there is [this issue](https://github.com/WordPress/gutenberg/issues/23011#issuecomment-1047766085) but I doubt anyone is actively working on it).

I think you'll need some kind of `has-external-link` class that can be added via Advanced (no need to make a block style), but then also combine it with the `:not(…)` code to avoid accidentally marking internal links if there are multiple in a paragraph, for example.

StevenDufresne: This is why I added a second `CSS` line to the declaration, `.external-link > a::after,` so it can be added to the parent. Not super excited about that but at least it works?

https://github.com/WordPress/wporg-parent-2021/pull/53/commits/1723230ccce77a0a8cbfdb2c41c4e1fc41c4923e#diff-dc9ae95ab0a8f0d1179f51200575b6479f49df7f5517f1afeef40a86aa475d09R32

StevenDufresne: > I think you'll need some kind of has-external-link class that can be added via Advanced (no need to make a block style), but then also combine it with the :not(…) code to avoid accidentally marking internal links if there are multiple in a paragraph, for example.

Ah originally misread that. Yeah, multiple links is an issue. Thanks for flagging that.

StevenDufresne:
Here's the updated view:

| Place | Image | 
| --- | --- |
| Regular link | Screen Shot 2022-12-08 at 4 19 03 PM |
| Scales with font size | Screen Shot 2022-12-08 at 4 24 13 PM |
| Embedded link | Screen Shot 2022-12-08 at 4 18 35 PM |
| Embedded link `light` | Screen Shot 2022-12-08 at 4 18 26 PM |

**Notes:**
- I have added a `external-link-light` for places where we have dark backgrounds
- I have also added styles in case we have access to the anchor tag directly

ryelle (approved):
Looks good 👍🏻

PR #54, opened 2022-11-17
Add 404 page

### LINKED ISSUE #15: Layout: 404 Page
Should look like [wordpress.org/news](https://wordpress.org/news/asdfasdf).

# Designs
Screen Shot 2022-11-01 at 10 14 52 AM

outdoor2kode:
Fixes https://github.com/WordPress/wporg-showcase-2022/issues/15

### Screenshots

| Desktop | Tablet | Mobile |
| --- | --- | ---|
| Screen Shot 2022-11-23 at 5 42 10 PM | Screen Shot 2022-11-23 at 5 42 23 PM | Screen Shot 2022-11-23 at 5 42 31 PM |

1. Set up a local env.
2. Go to a random page, for example http://localhost:8898/random.
3. Check the 404 page.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Title: 404 Page Subtitle
+ * Slug: wporg-parent-2021/404-page-subtitle
+ * Categories: wporg
+ */
+
+?>
+
+<!-- wp:paragraph -->
+<p>
+	<?php esc_html_e( 'Go to', 'wporg' ); ?>
+	<a href="https://wordpress.org"><?php esc_html_e( 'the homepage', 'wporg' ); ?></a>,
+	<?php esc_html_e( 'or try searching WordPress.org sites using the field below.', 'wporg' ); ?>
+</p>
+<!-- /wp:paragraph -->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
@@ -2,15 +2,19 @@
 /**
  * Title: 404 Page Subtitle
  * Slug: wporg-parent-2021/404-page-subtitle
- * Categories: wporg
+ * Inserter: no
  */
 
 ?>
 
 <!-- wp:paragraph -->
 <p>
-	<?php esc_html_e( 'Go to', 'wporg' ); ?>
-	<a href="https://wordpress.org"><?php esc_html_e( 'the homepage', 'wporg' ); ?></a>,
-	<?php esc_html_e( 'or try searching WordPress.org sites using the field below.', 'wporg' ); ?>
+	<?php
+	printf(
+		/* translators: %s is the site URL. */
+		wp_kses_post( __( 'Go to <a href="%s">the homepage</a> or try searching WordPress.org sites using the field below.', 'wporg' ) ),
+		esc_url( get_site_url( '/' ) )
+	);
+	?>
 </p>
 <!-- /wp:paragraph -->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
@@ -12,7 +12,7 @@
 	<?php
 	printf(
 		/* translators: %s is the site URL. */
-		wp_kses_post( __( 'Go to <a href="%s">the homepage</a> or try searching WordPress.org sites using the field below.', 'wporg' ) ),
+		wp_kses_post( __( 'Go to <a href="%s">the homepage</a> or try searching using the field below.', 'wporg' ) ),
 		esc_url( get_site_url( '/' ) )
 	);
 	?>
--- a/source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
@@ -15,7 +15,7 @@
 <!-- /wp:group -->
 
 <!-- wp:heading {"level":1} -->
-<h1><?php esc_html_e( 'This page doesn\'t exist.', 'wporg' ); ?></h1>
+<h1><?php esc_html_e( 'This page doesn&#146;t exist.', 'wporg' ); ?></h1>
 <!-- /wp:heading -->
 
 <!-- wp:pattern {"slug":"wporg-parent-2021/404-page-subtitle"} /-->
--- a/source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
+++ b/source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
@@ -2,7 +2,7 @@
 /**
  * Title: 404 Page
  * Slug: wporg-parent-2021/404-page
- * Categories: wporg
+ * Inserter: no
  */
 
 ?>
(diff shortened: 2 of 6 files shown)


source/wp-content/themes/wporg-parent-2021/patterns/404-page-subtitle.php
```diff
@@ -0,0 +1,16 @@
...
+ * Categories: wporg
+ */
+
+?>
+
+<!-- wp:paragraph -->
+<p>
+	<?php esc_html_e( 'Go to', 'wporg' ); ?>
+	<a href="https://wordpress.org"><?php esc_html_e( 'the homepage', 'wporg' ); ?></a>,
+	<?php esc_html_e( 'or try searching WordPress.org sites using the field below.', 'wporg' ); ?>
```
ryelle: Since this is one sentence, it should be all in the same translation call. HTML can be used & sanitized with `wp_kses_post`. Also, maybe we can use `get_site_url` instead of hardcoding wp.org, which should return the user to the homepage of the site they're currently on.

```suggestion
	<?php
	printf(
		/* translators: %s is the site URL. */
		wp_kses_post( __( 'Go to <a href="%s">the homepage</a> or try searching WordPress.org sites using the field below.', 'wporg' ) ),
		esc_url( get_site_url( '/' ) )
	);
	?>
```

source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Title: 404 Page
+ * Slug: wporg-parent-2021/404-page
+ * Categories: wporg
```
ryelle: This isn't a pattern meant to be used in content, so it doesn't need a category. Instead, we can disable it from the editor using:
```suggestion
 * Inserter: no
```

source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+?>
+
+<!-- wp:group {"className":"site-content-container"} -->
+<div class="wp-block-group site-content-container">
+	<p class="error404__oops" aria-hidden="true">
+		<?php esc_html_e( 'Oops!', 'wporg' ); ?>
+	</p>
```
ryelle: This should be in a block wrapper… but maybe it should actually be added in via CSS? Currently, since it's plain text, it can be highlighted:

Screen Shot 2022-11-17 at 16 15 10

Adding it in CSS would loose the ability to translate it, but that's okay for now.

source/wp-content/themes/wporg-parent-2021/theme.json
```diff
@@ -121,7 +121,9 @@
 		},
 		"custom": {
 			"alignment": {
-				"alignedMaxWidth": "50%"
+				"alignedMaxWidth": "50%",
+				"//edge-spacing": "The min and max values represent the mobile and desktop spacing; the calc() value is for tablets.",
+				"edge-spacing": "clamp( 24px, calc( 100vw / 18 ), 80px )"
```
ryelle: Rather than adding this value, can you use `var(--wp--preset--spacing--60)`? I've been trying to get us to stick to [the defined spacing values](https://github.com/WordPress/wporg-parent-2021/#spacing), and that one is `clamp( 20px, calc( 100vw / 18 ), 80px )`.

Even the global header (which uses that value) will be changing soon (https://github.com/WordPress/wporg-mu-plugins/issues/301).

ryelle (changes requested):

ryelle:
Where did the design for this page come from? I can't find a 404 page design for showcase, and I'm wondering about the Oops color (it looks darker than I would expect). I also think this should just be using the global header & footer (rather than the template parts) to better match the news design.

StevenDufresne:
> Where did the design for this page come from? I can't find a 404 page design for showcase, and I'm wondering about the Oops color (it looks darker than I would expect). I also think this should just be using the global header & footer (rather than the template parts) to better match the news design.

There isn't a design, but an agreement to use the same one as news. I think if we ship with the dark version would equally be ok, although a white version aligns better with the theme. I agree the oops text can be lightened.

ryelle:
> There isn't a design, but an agreement to use the same one as news. I think if we ship with the dark version would equally be ok, although a white version aligns better with the theme. I agree the oops text can be lightened.

@jasmussen Can you chime in with how a light 404 page should look? Some screenshots in the description above of this current implementation.

jasmussen:
Ah, thanks for the ping.

My personal take is that the fewer different 404 pages we can have, the better, i.e. it would be great if the 404 for all (most?) subpages literally were the same, there doesn't seem to be a great reason to fragment here.

However if there is a need for a light version, I would put it as close to https://github.com/WordPress/wporg-showcase-2022/issues/15 in inverted form as possible. That is, the "Oops" is entirely decorative, and should be as light a gray as we can muster. There also shouldn't be any header or subnavigation. CC: @beafialho @javierarce

bengreeley:
I agree, I'd love it if the 404 pages were as universal as possible across the website.

What are the plans for the 'Notes' and 'Call to Action' sections on the bottom? We'll want to be careful about caching any database queries on 404 pages since it's going to get so much traffic at so many different URLs. This is probably something you all are already considering, but I wanted to be sure I mentioned it. :-) I wonder if we need those sections and if we should simply match the simplicity of the [News 404 page](https://wordpress.org/news/dkdkd).

Thank you for your work on this!

outdoor2kode:
Thanks for the feedback! @bengreeley @jasmussen 

The header, subnav, and the 'Notes' and 'Call to Action' sections have been removed. 
It has also been changed to the dark version, the style is based on the [News 404 page](https://wordpress.org/news/dkdkd).
 I've updated the screenshots in the description.

outdoor2kode:
@jasmussen May I ask your thought about the text in the red square in the attached img? Do you think we can get rid of the text and just change it to something like `try searching sites using the field below.`? This allows us not to create a new pattern in a child theme (e.g., showcase) simply for overriding the text. In News, the search only searches locally rather than all of the .org, so it seems to me that it's ok to do so.

image

jasmussen:
For the sake of going generic, I think we could go even simpler, yes. I'd be happy just with:

> Go to the homepage or try searching using the field below.

source/wp-content/themes/wporg-parent-2021/patterns/404-page.php
```diff
@@ -0,0 +1,24 @@
...
+<main class="wp-block-group site-content-container has-white-color has-charcoal-2-background-color has-text-color has-background" style="border-top-color:var(--wp--preset--color--charcoal-1);border-top-width:1px;border-right-style:none;border-right-width:0px;border-bottom-style:none;border-bottom-width:0px;border-left-style:none;border-left-width:0px;margin-top:0;padding-right:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"textColor":"charcoal-1","className":"wporg-parent-oops-container","layout":{"type":"constrained"}} -->
+<div class="wp-block-group wporg-parent-oops-container has-charcoal-1-color has-text-color"><!-- wp:paragraph {"textColor":"white"} -->
+<p class="has-white-color has-text-color">Change the text color of the group here to adjust "Oops" color.</p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group -->
+
+<!-- wp:heading {"level":1} -->
+<h1><?php esc_html_e( 'This page doesn\'t exist.', 'wporg' ); ?></h1>
```
ryelle: This should be using a curly quote since it's in human-facing text.

News:
Screenshot 2022-11-29 at 11 42 16 AM

This PR:
Screenshot 2022-11-29 at 11 42 12 AM

Can be fixed with this:
```suggestion
<h1><?php esc_html_e( 'This page doesn&#146;t exist.', 'wporg' ); ?></h1>
```

ryelle (approved):
One suggestion, otherwise this looks good 💯

PR #60, opened 2022-11-28
Fix dropdown layout broken issue

outdoor2kode:
See https://github.com/WordPress/wporg-showcase-2022/pull/64.

### Screenshots
dropdown

1. Go to random showcase single page
2. Switch to mobile view
3. Test dropdown

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -479,9 +479,10 @@
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ryelle:
@renintw I split up the styles a little more (in https://github.com/WordPress/wporg-parent-2021/pull/60/commits/10aa85b4d2c402147ddc254e58fb606ab3d4f34f), so now the `is-style-brush-stroke` class should only apply a brush stroke, while the `is-sticky` class triggers the sticky behavior. That should make it so that showcase only needs to remove the `is-brush-stroke-style` class (and fix the padding) but not need any CSS changes. And the download subpages can keep both the brush stroke and sticky classes for a sticky header with a brush stroke.

PR #61, opened 2022-11-30
Block Styles: Add the Secondary Search block style on search block

ryelle:
The secondary search style is needed for the documentation site as well as showcase, so this PR moves the code from showcase to the parent theme.

I made a few changes to the styles in converting the CSS to Sass, the biggest change was remove `!important` when it wasn't necessary, and setting the custom properties for the button colors so the interactive states still work.

Lastly, I added a local custom prop for the border color, since documentation uses a light grey border, this way it can be set like this, to apply to all the borders.

```css
.wp-block-search.is-style-secondary-search-control {
	--local--border-color: var(--wp--preset--color--light-grey-1);
}
```

### Screenshots

Style is available on any site using the parent theme

Screenshot 2022-11-30 at 1 54 36 PM

The variations of the search block with Secondary style

Screenshot 2022-11-30 at 1 51 12 PM

In context on showcase

Screenshot 2022-11-30 at 1 41 27 PM

1. Add a search block to a page or template, you should see the Secondary style option
2. Apply the style, the block should match the screenshots above (the editor view doesn't totally match, check frontend only)
3. Try with the showcase PR TBD, there should be no visual change on showcase.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -148,6 +148,15 @@ function setup_block_styles() {
 			'style_handle' => STYLE_HANDLE,
 		)
 	);
+
+	register_block_style(
+		'core/search',
+		array(
+			'name'         => 'secondary-search-control',
+			'label'        => __( 'Secondary', 'wporg' ),
+			'style_handle' => STYLE_HANDLE,
+		)
+	);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


StevenDufresne (approved):
While I didn't test all the options, the search form on showcase is still working as it did before 👍

StevenDufresne:
We should transfer over https://github.com/WordPress/wporg-parent-2021/issues/62 when this merges.

outdoor2kode (approved):
Tested on showcase and parent. Looks good to me 👍

PR #66, opened 2022-12-19
Use currentColor for the external link.

StevenDufresne:
We have a case where we need the [external link to be blueberry](https://github.com/WordPress/wporg-showcase-2022/pull/110#issuecomment-1354572161). Instead of adding new classes for all the colors we could make use of `currentcolor` and use the text color. I initially missed that in the first PR. I wasn't aware that it would work in this case.
 
## Considerations
I don't believe there are any uses of `external-link` in the wild that would need replacing so this should be straightforward.

### Screenshots
|  | IMG |
| --- | --- |
| Color | Screen Shot 2022-12-19 at 10 37 42 AM |
| Light| Screen Shot 2022-12-19 at 10 37 22 AM|
| Dark | Screen Shot 2022-12-19 at 10 39 59 AM|

1. Load this PR
2. Add the class `external-link` to any `<a>` tag.
3. Expect the icon to have the color of the `<a>` tag.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -30,23 +30,16 @@ mark.has-inline-color {
 
 // For links that navigate users to different site
 a.external-link::after,
-a.external-link-light::after,
-.external-link > a:not([href*="wordpress.org"])::after,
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Works for me 🎉 

Screen Shot 2022-12-19 at 3 00 11 PM

PR #67, opened 2023-01-09
Styles: Update caption styles on gallery, images

ryelle:
This updates the gallery caption styles to better match the regular image & other media captions. It also centers the caption for all media — originally the captions were defaulted to right alignment to match News ([example](https://wordpress.org/news/2022/10/people-of-wordpress-raghavendra-satish-peri/)), but this was changed to centered for the Documentation design. @javierarce can you confirm whether centered captions should be the default for all the redesigned sites?

Fixes https://github.com/WordPress/wporg-documentation-2022/issues/29

### Screenshots

| | Before | After |
|----|--------|-------|
| Single image | before-image-w-caption | after-image-w-caption |
| Gallery | before-gallery-w-captions | after-gallery-w-captions |

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -52,20 +52,28 @@ figure[class*="wp-block-"],
 [class*="wp-block-"] {
 	figcaption,
 	.wp-element-caption {
-		text-align: right;
+		text-align: center;
 	}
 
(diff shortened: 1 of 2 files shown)


javierarce:
> @javierarce can you confirm whether centered captions should be the default for all the redesigned sites?

In my opinion, this should be the default style for all the redesigns, but News could keep the current right-side alignment style.

cc: @beafialho @jasmussen @panchovm

ryelle:
News isn't part of this system, so it will keep the right alignment 👍🏻

PR #72, opened 2023-02-01
Add list block style for long items

### LINKED ISSUE #71: Add spaced out block style for lists with long format items
The redesigned About/Features page has list items with long format content which require more spacing than normal:

Screen Shot 2023-02-02 at 12 27 04 PM

It has been [suggested that we should add a new block style](https://github.com/WordPress/wporg-main-2022/pull/176#discussion_r1093613819), as the list block doesn't use block gap for this.

adamwoodnz:
Adds a 'Long items' style to list blocks which spaces out the list items. Required for spacing of content in the redesigned About/Features page, see #71 for design.

Closes #71 

Props @ryelle 

### Screenshots

| Editor | Frontend |
|--------|-------|
| Screen Shot 2023-02-02 at 12 37 10 PM | localhost_8888_about-font-test_(Desktop) |

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
@@ -585,3 +585,7 @@
 		padding: var(--wp--preset--spacing--10) !important;
 	}
 }
+
+.is-style-list-long-items > li {
+	margin-bottom: var(--wp--preset--spacing--20);
+}
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -141,6 +141,15 @@ function setup_block_styles() {
 	);
 
 	register_block_style(
+		'core/list',
+		array(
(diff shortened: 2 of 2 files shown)


source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss
```diff
@@ -585,3 +585,7 @@
 		padding: var(--wp--preset--spacing--10) !important;
 	}
 }
+
+.is-style-list-long-items > li {
+	margin-bottom: var(--wp--preset--spacing--30);
```
ryelle: The regular paragraph spacing is [20px](https://github.com/WordPress/wporg-parent-2021/blob/2b7ee519917e808de71e17e424e3a29287497e70/source/wp-content/themes/wporg-parent-2021/theme.json#L791), so I think `--wp--preset--spacing--20` would be more consistent here. I see the design uses 30px between paragraphs, but that doesn't match the other site designs.

adamwoodnz: Updated!

PR #73, opened 2023-02-07
Try using a filter to merge child & parent theme.json settings

ryelle:
This updates the theme.json resolver to merge in parent theme settings for the array properties (color palette, font sizes, etc) before merging back into the parent theme.

Previously, defining an array property would totally overwrite the parent theme value. Now this hooks in `wp_theme_json_data_theme` to merge the list defined in the parent theme into the child theme's value, using a helper function `_merge_by_slug`. This only updates those properties that need merging, the core process of merging the child & parent theme.json happens [right after this filter returns](https://github.com/WordPress/gutenberg/blob/53fab106f062458f5da0efa2f031184b9ce0381f/lib/class-wp-theme-json-resolver-gutenberg.php#L260-L275).

If the child theme.json has values with the same slug as something in the parent theme.json, the child theme will overwrite that single value. Additional values will be added to the end of the parent's list.

### Screenshots

There should be no visible changes on the current themes.

I've pushed a [testing branch](https://github.com/WordPress/wporg-developer/tree/try/local-style-changes) to developer, which adds a new color, updates some font sizes, and heading font using the latest i3 design. If you apply both and spin up developer, you'll see that the headings are all IBM Plex Mono. I also added green paragraphs to show that the color is set correctly 🙂 

In the editor, there are different font sizes for the headings (these overwrote the parent).

fontsizes

And the color palette has an additional green at the end (this was added to the parent).

colors

1. Spin up a child theme site.
2. Add some changes into one of the array value properties, `color.palette`, `typography.fontFamilies`, `spacing.spacingSizes`, etc.
3. Open the editor, you should see your updated settings.
4. Try editing other things in the child theme.json, it should still update correctly on the front end.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/functions.php
+++ b/source/wp-content/themes/wporg-parent-2021/functions.php
@@ -17,6 +17,7 @@ add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_filter( 'author_link', __NAMESPACE__ . '\use_wporg_profile_for_author_link', 10, 3 );
 add_filter( 'render_block_core/pattern', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
 add_filter( 'the_content', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
+add_filter( 'wp_theme_json_data_theme', __NAMESPACE__ . '\merge_parent_child_theme_json' );
 
 /**
  * Register theme support.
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-parent-2021/functions.php
```diff
@@ -106,6 +107,114 @@ function prevent_arrow_emoji( $content ) {
...
+		$parent_theme           = new \WP_Theme_JSON_Gutenberg( $parent_theme_json_data );
+
+		// Get base theme.json settings.
+		$theme_settings = $parent_theme->get_settings();
+		$settings       = $child_theme['settings'];
+
+		// Define the array values here, so they can be updated if the theme.json schema changes.
+		$color_keys = [ 'duotone', 'gradient', 'palette' ];
+		$typog_keys = [ 'fontFamilies', 'fontSizes' ];
+		$space_keys = [ 'spacingSizes' ];
```
StevenDufresne, line 151: Ideally, we wouldn't have to define these keys but I think it's _good enough_ for what we need right now. 

Also in an ideal world, core/Gutenberg handles this logic.

StevenDufresne (approved):
I tested out the branch. The changes appear to be working! 

It would be nice to have some test coverage for these functions although it's not a requirement.

source/wp-content/themes/wporg-parent-2021/functions.php
ryelle: > Ideally, we wouldn't have to define these keys but I think it's good enough for what we need right now.

We could probably build up which properties are "array" by traversing the schema, but by defining them like this we also know that this set definitely all have `slug` properties (there is an array with no slug, spacing.units, and that is not handled here). Anyway, it seemed easier to hardcode these than try to build them dynamically. I think the schema has been fairly stable since it was versioned 2.

> Also in an ideal world, core/Gutenberg handles this logic.

Agreed, but there was enough back and forth on [the GB issue](https://github.com/WordPress/gutenberg/issues/40557) about whether this should be the case that I don't see it happening soon.

source/wp-content/themes/wporg-parent-2021/functions.php
```diff
@@ -106,6 +107,114 @@ function prevent_arrow_emoji( $content ) {
...
+	$child_theme = $theme_json->get_data();
+
+	if ( ! empty( $child_theme['settings'] ) ) {
+		$parent_theme_json_file = get_template_directory() . '/theme.json';
+		$parent_theme_json_data = wp_json_file_decode( $parent_theme_json_file, array( 'associative' => true ) );
+		$parent_theme           = new \WP_Theme_JSON_Gutenberg( $parent_theme_json_data );
+
+		// Get base theme.json settings.
+		$theme_settings = $parent_theme->get_settings();
+		$settings       = $child_theme['settings'];
```
adamwoodnz: I think it would make it clearer further down if these were more specifically named
```suggestion
		$parent_settings = $parent_theme->get_settings();
		$child_settings  = $child_theme['settings'];
```

adamwoodnz (approved):
Nice! This is great, just one minor suggestion.

source/wp-content/themes/wporg-parent-2021/functions.php
ryelle: Good catch — my first try at this used style variations, so it was theme settings vs a given variation, and I didn't change all the variables. Updated in https://github.com/WordPress/wporg-parent-2021/pull/73/commits/d607c11e976e7ac518cd5784d0c9e87025f8f1f4.

adamwoodnz (approved):
Tested in wporg-main-2022 with the extra font and works for me 🎉

PR #77, opened 2023-03-21
Increase font-size of table text to match body

### LINKED ISSUE #76: Default table style
The default table style uses a font size which is too small:

- Bump the font size of jazzers inside the [WordPress history release table](https://wordpress.org/about/history/). It should be at least the same font size as the body text above.
- If we use tables like this, with a smaller font size inside, elsewhere on the site, we should bump those as well.

History page
Screen Shot 2023-03-22 at 10 13 06 AM

Roadmap page
Screen Shot 2023-03-22 at 10 17 43 AM

adamwoodnz:
See #76 

### Screenshots

| Before | After |
|--------|-------|
| Screen Shot 2023-03-22 at 11 54 24 AM | Screen Shot 2023-03-22 at 11 56 01 AM |

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_table.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_table.scss
@@ -9,6 +9,10 @@
 }
 
 .wp-block-table {
+	> table {
+		font-size: var(--wp--preset--font-size--normal);
+	}
+
 	th {
 		text-align: left;
 		font-weight: 400;


ryelle:
@adamwoodnz The font size here is actually controlled [in theme.json](https://github.com/WordPress/wporg-parent-2021/blob/19744d55fd7d682c530c1e99ebec3b73c79b8631/source/wp-content/themes/wporg-parent-2021/theme.json#L701-L705) — I think if you remove that setting, you probably don't need the CSS… or maybe update the setting to `normal`?

adamwoodnz:
> @adamwoodnz The font size here is actually controlled [in theme.json](https://github.com/WordPress/wporg-parent-2021/blob/19744d55fd7d682c530c1e99ebec3b73c79b8631/source/wp-content/themes/wporg-parent-2021/theme.json#L701-L705) — I think if you remove that setting, you probably don't need the CSS… or maybe update the setting to `normal`?

Aargh good point, I'll update, thanks.

PR #81, opened 2023-03-29
Remove text highlighting to avoid confusion

### LINKED ISSUE #80: Misc: Revisit the text highlight color
At the moment, selecting text and items highlights in red/orange/coral, which has an appearance of being a warning:

Screenshot 2023-03-28 at 14 28 05

Additionally, it doesn't invert the links, and is not quite as flexible as the native highlight color. Let's remove the custom highlight for now.

iandunn:
Fixes #80

Screenshot 2023-03-28 at 5 56 15 PM


jasmussen:
Yep let's go with that for now. If we find a solid pattern for the selection customization, we can always revisit.

jasmussen (approved):

PR #83, opened 2023-04-11
Element styles: Add margin to h2 & h3

ryelle:
This moves the heading margin from `theme.json`, where it only applied to real content (articles in docs), but not content from patterns (the page content on the main site). Now the margin is applied to all `h2`s and `h3`s without custom font sizes (if a font size is added and a margin is necessary, it can be added in the editor too).

See https://github.com/WordPress/wporg-main-2022/issues/246
This is an alternative to https://github.com/WordPress/wporg-main-2022/pull/251, assuming these values should cascade to all pages on the site. If the margins should only apply to the current pages, the in-content approach of 251 makes more sense, but since the same values are used on Documentation, I thought it should be provided by the parent theme.

### Screenshots

Ran `blink-diff` across screenshots. The first screenshot is `trunk`, the last is this PR, and the middle highlights differences. The only difference is the spacing on About - Philosophy & About - Requirements.

result-home-1280
result-download-1280
result-dl-releases-1280

result-about-1280
result-about-philosophy-1280
result-about-requirements-1280

result-docs-home-1280
result-docs-topic-landing-1280
result-docs-category-1280
result-docs-article-1280

1. Pull down the branch and build the styles `yarn workspaces run build`
2. Spin up a child theme site (see [step 5 in docs setup](https://github.com/WordPress/wporg-documentation-2022#setup) or [main setup](https://github.com/WordPress/wporg-main-2022/#setup) for testing with a real child theme) or push to a sandbox.
3. Any pages with plain h2 & h3 should have spacing, for example articles on Documentation or pages in the About section
4. The spacing does not affect other existing pages (generally because they've got custom font sizes)

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -125,6 +125,16 @@ input[type="checkbox"] + label {
 	line-height: 1em; // stylelint-disable-line declaration-property-unit-allowed-list
 }
 
+// Headings.
+// Add margin to default h2 & h3 (no font size changes).
+.wp-site-blocks h2:not([class*="-font-size"], [style*="font-size"]) {
+	margin-top: var(--wp--preset--spacing--40);
+}
+
+.wp-site-blocks h3:not([class*="-font-size"], [style*="font-size"]) {
+	margin-top: var(--wp--preset--spacing--30);
+}
+
 // Line heights for custom sizes.
 .has-extra-small-font-size {
 	line-height: var(--wp--custom--body--extra-small--typography--line-height);
--- a/source/wp-content/themes/wporg-parent-2021/theme.json
+++ b/source/wp-content/themes/wporg-parent-2021/theme.json
@@ -642,24 +642,6 @@
 					"fontSize": "var(--wp--preset--font-size--normal)"
 				}
 			},
-			"core/post-content": {
-				"elements": {
-					"h2": {
-						"spacing": {
-							"margin": {
-								"top": "var(--wp--preset--spacing--40)"
-							}
-						}
-					},
-					"h3": {
-						"spacing": {
-							"margin": {
-								"top": "var(--wp--preset--spacing--30)"
-							}
-						}
-					}
-				}
-			},
 			"core/post-date": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--small)"


source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
```diff
@@ -125,6 +125,16 @@ input[type="checkbox"] + label {
 	line-height: 1em; // stylelint-disable-line declaration-property-unit-allowed-list
 }
 
+// Headings.
+// Add margin to default h2 & h3 (no font size changes).
+.wp-site-blocks h2:not([class*="-font-size"], [style*="font-size"]) {
```
StevenDufresne, line 130: I can't remember off the top of my head, but do we control the class `.wp-site-blocks`? It would be nice to know that this can't be removed or renamed by gutenberg and break our styling.

StevenDufresne: I'm also a bit wary of using Gutenberg styles in the `:not` clause although I can't think of a better solution. 🤔

source/wp-content/themes/wporg-parent-2021/theme.json
```diff
@@ -642,24 +642,6 @@
 					"fontSize": "var(--wp--preset--font-size--normal)"
 				}
 			},
-			"core/post-content": {
```
outdoor2kode, line 645: I was hoping `core/pattern` could have some effects on `pattern content` like `core/post-content` here does to `post content`, but it doesn't seem to work that way 🤔

ryelle: Yeah, that was my first attempt too 😞  I think because `pattern` isn't a normal block - it doesn't output any wrapper around the pattern content.

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
ryelle: We don't control `.wp-site-blocks`, but we're already using it for some other basic layout styles, so if it disappears we'll know pretty quickly. This class is also being used in at least Twenty Twenty-Two, so we're not going totally against the grain. The issue we ran into with the other container class was that it was a generated class, so GB wasn't expecting anyone to use it.

> I'm also a bit wary of using Gutenberg styles in the :not clause although I can't think of a better solution. 🤔

I think we can be fairly confident these aren't going to change, though. I'm pretty sure it's been `has-*-font-size` since that feature started, and `style="font-size…"` when it's a custom size seems unlikely to change.

source/wp-content/themes/wporg-parent-2021/theme.json
outdoor2kode: It appears so, there's no wrapper like `wp-block-post-content` around the pattern. Not sure if GB designed it this way on purpose, or just missed it.

outdoor2kode (approved):
Tested and looks good, `blink-diff` is awesome 💯

PR #84, opened 2023-04-11
Fix content breaking out of main container

ryelle:
Fixes https://github.com/WordPress/wporg-documentation-2022/issues/50 — Force a word break on content so that the text wraps rather than breaking out of the container.

This also removes the wrap from `pre` blocks, since they're usually code and the wrap distorts the indentation.

### Screenshots

Three examples from Docs, and one from About, to show this works across the child themes.

| | Before | After |
|---|---|---|
| [1. Preformatted](https://wordpress.org/documentation/article/tools-network-screen/)<br />Docs | t-preformatted | b-preformatted |
| [2. Code](https://wordpress.org/documentation/article/wordpress-glossary/)<br />Docs | t-list-code | b-list-code |
| [3. Link](https://wordpress.org/about/privacy/)<br />About | t-list-link | b-list-link |
| [4. Tables](https://wordpress.org/documentation/article/comparing-patterns-template-parts-and-reusable-blocks/)<br />No difference, already works | t-table | b-table |

https://wordpress.org/documentation/article/twenty-twenty-changelog/ — For this one I edited the article itself to remove the `&nbsp;`s between the words, which along with the break-word rule correctly wraps on small screens.

https://wordpress.org/documentation/article/what-are-smilies/ — This was another page with slightly malformed content, the table block was missing the figure wrapper. Just by updating the page I fixed the table display and now it overflows with a scroll instead of breaking the container.

1. Apply this to a sandbox
2. View any of the linked pages on https://github.com/WordPress/wporg-documentation-2022/issues/50
3. The page should not have a horizontal scroll
4. Some elements on the page (preformatted blocks, tables) might have a horizontal scroll
5. Check other pages at larger sizes too for regressions

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_preformatted.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_preformatted.scss
@@ -1,0 +1,5 @@
+.wp-block-preformatted {
+	white-space: pre;
+	overflow-x: scroll;
+	overflow-wrap: normal;
+}
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
@@ -16,7 +16,6 @@
 	margin-top: var(--wp--custom--margin--baseline);
 
 	> label {
-		width: 100%;
 		margin-bottom: calc(var(--wp--custom--form--padding) / 2);
 		font-size: var(--wp--custom--form--typography--font-size);
 		color: var(--wp--custom--form--color--label);
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
@@ -3,6 +3,7 @@
 	display: flex;
 	flex-direction: column;
 	min-height: calc(100vh - var(--wp-global-header-offset, 0px) - var(--local-header-height, 0px));
+	overflow-wrap: break-word;
 
 	> main,
 	> article {
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_index.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_index.scss
@@ -8,6 +8,7 @@
 @import "post-excerpt";
 @import "post-navigation-link";
 @import "post-title";
+@import "preformatted";
 @import "pullquote";
 @import "query-pagination";
 @import "quote";


source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
```diff
@@ -16,7 +16,6 @@
 	margin-top: var(--wp--custom--margin--baseline);
 
 	> label {
-		width: 100%;
```
StevenDufresne, line 19: Is this in the right PR?

source/wp-content/themes/wporg-parent-2021/sass/blocks/_preformatted.scss
```diff
@@ -0,0 +1,5 @@
+.wp-block-preformatted {
+	white-space: pre;
```
StevenDufresne, line 2: This is causing an issue:

[Video Shortcode](https://wordpress.org/documentation/article/video-shortcode/):

Screenshot 2023-04-13 at 1 01 19 PM

StevenDufresne:
I'm still getting the issue I [commented on](https://github.com/WordPress/wporg-documentation-2022/issues/50#issuecomment-1503606783) on [this page ](http://localhost:8888/article/how-to-install-wordpress/#step-3-set-up-wp-config-php). Should we address this here? Its probably likely that other examples exist.

source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
ryelle: Yeah, this causes the content to break out of the screen when the search is in a sidebar.

ryelle:
> I'm still getting the issue I https://github.com/WordPress/wporg-documentation-2022/issues/50#issuecomment-1503606783 on [this page ](http://localhost:8888/article/how-to-install-wordpress/#step-3-set-up-wp-config-php). Should we address this here? Its probably likely that other examples exist.

I didn't fix that because that page is unpublished, no one will see it (in fact it should redirect to https://developer.wordpress.org/advanced-administration/before-install/howto-install/ now). The cause of the issue was the content itself- the image was copied over from maybe the Codex? with extra styles/classes. It's the only article I can find using `thumbinner` or `thumbimage`, so I think it's fine.

source/wp-content/themes/wporg-parent-2021/sass/blocks/_preformatted.scss
ryelle: I've fixed the content there — looks like it was hit by [this shortcode issue](https://core.trac.wordpress.org/ticket/55996#comment:29), it was not meant to actually embed any videos. If a video is embedded not in a `pre`, it should wrap the error message just fine.

StevenDufresne (approved):
From what I tested works well. 👍  

We can open new tickets for more example that are found.

PR #87, opened 2023-06-12
Styles: Add underline to default link style.

ryelle:
I noticed in https://github.com/WordPress/wporg-main-2022/issues/268#issuecomment-1586557246 that the design appears to have underlines on links by default, but our parent theme removes those. Links in post content are underlined, but in the main theme we don't use post content blocks. On Documentation, there's a rule to [add underlines](https://github.com/WordPress/wporg-documentation-2022/blob/8b48a2645a2b080222b2fea56e178ff68d13f84b/source/wp-content/themes/wporg-documentation-2022/src/style/style.scss#L11) to keep things consistent, but if this is merged we can probably remove that.

This only affects default link styles — buttons, links in navigation menus, and styled links (like the list pattern) are not affected.

### Screenshots

Note that the homepage before screenshot shows the banner link not-underlined, that's because I actually added a manual underline to it on production, which was removed for testing this PR.

| Before | After |
|--------|-------|
| before-home | after-home |
| before-download | after-download |
| before-about | after-about |
| before-about-reqs | after-about-reqs |
| before-docs | after-docs |

1. Spin up any of the child theme sites (main, documentation, etc)
2. Check that default links have underlines, and that the underlines disappear on hover.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -1,14 +1,13 @@
 // Links
 a {
 	cursor: pointer;
-	text-decoration: none;
 	text-decoration-thickness: 1px;
 	text-underline-offset: 0.15em;
 	font-weight: revert; // stylelint-disable-line font-weight-notation
 
 	&:hover,
 	&:focus {
-		text-decoration-line: underline;
+		text-decoration-line: none;
 	}
 }
 
--- a/source/wp-content/themes/wporg-parent-2021/theme.json
+++ b/source/wp-content/themes/wporg-parent-2021/theme.json
@@ -647,6 +647,20 @@
 					"fontSize": "var(--wp--preset--font-size--small)"
 				}
 			},
+			"core/post-navigation-link": {
+				"elements": {
+					"link": {
+						"typography": {
+							"textDecoration": "none"
+						},
+						":hover": {
+							"typography": {
+								"textDecoration": "underline"
+							}
+						}
+					}
+				}
+			},
 			"core/post-terms": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--small)"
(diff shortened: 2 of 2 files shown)


StevenDufresne:
On the downloads page, [Releases](https://wordpress.org/download/releases/), [Nightly](https://wordpress.org/download/beta-nightly/), [Counter](https://wordpress.org/download/counter/), [Source](https://wordpress.org/download/source/) are not underlined. Should they be? [Relevant code](https://github.com/WordPress/wporg-parent-2021/blob/68e7ed543d3f00754fd164746909637a0cb9e97e/source/wp-content/themes/wporg-parent-2021/sass/blocks/_navigation.scss#L4).

jasmussen:
Let's find the balance that is consistent, I recognize there's a lack of a system so I'll make sure we discuss this in a chat too. E.g. to @StevenDufresne's point. I.e. these 4 blue links should remain as they are:

Screenshot 2023-06-13 at 11 13 50

Same with breadcrumbs in the incoming designs, where you'll note some of the list items are also not underlined:
Screenshot 2023-06-13 at 11 16 00

I'll return!

ryelle:
> On the downloads page, [Releases](https://wordpress.org/download/releases/), [Nightly](https://wordpress.org/download/beta-nightly/), [Counter](https://wordpress.org/download/counter/), [Source](https://wordpress.org/download/source/) are not underlined. Should they be?
> these 4 blue links should remain as they are:

They are not underlined, because navigation menu links remove the underlines (it's a core block style). You can see in the ["after" screenshot](https://user-images.githubusercontent.com/541093/245224305-02530a33-7b10-479f-9b53-42b41c96f853.png) that they are not underlined.

> Same with breadcrumbs in the incoming designs, where you'll note some of the list items are also not underlined:

I think we're good, as this is the _default_ state. In these other cases (nav menus, breadcrumbs, etc) we can override that because there are other contextual clues that this is a link. For example, if there was a link in that paragraph under Tutorials, it should have an underline, to differentiate it from the rest of the text (because the blueberry/charcoal contrast is only 2.9, and it needs to be 3.0+ if color alone).

StevenDufresne (approved):
Looks good to me. We can address one offs in different context as they come up.

👍

ryelle:
I proactively removed links from the post navigation & site title in theme.json here, and after merging this I've also got a commit for wporg-mu-plugins to remove it from the breadcrumbs & table of contents.

PR #94, opened 2023-08-15
Prevent screen readers from reading the arrow emoji in content

ryelle:
See https://github.com/WordPress/wporg-main-2022/issues/311

> Please eliminate the "↗︎" icons on the links under the "Only the beginning" heading 2. You never know what screen readers will do when they come in contact with arrows. In all fairness, up arrows really don't make much sense here.

This PR will wrap any arrows in content (or patterns, which are content) in a `<span aria-hidden="true">` tag. There should be no visual change, but it should improve the screen reader experience. Adding this to the parent theme will cascade the effect down to all child themes, which share this links list with arrows pattern.

### Screenshots

Confirming no visual change.

| Before | After |
|--------|-------|
| trunk-home | branch-home |
| trunk-blocks | branch-blocks |

1. Apply the branch, and start an environment using a child theme, for example the main site
2. View a page with arrows (home, blocks) or create a new page with arrow emoji — use any of `←↑→↓↔↕↖↗↘↙`
3. When viewed on the frontend, the arrows should be wrapped in span


StevenDufresne (approved):
LGTM

adamwoodnz (approved):
👍 Tested in NVDA and VoiceOver

PR #98, opened 2023-08-21
404: Pass correct `get_site_url` param to avoid infinite loop

iandunn:
The first param to `get_site_url()` should be a site ID, or `null` for the current site. Passing a string can cause a `switch_to_blog( 0 )` in some circumstances. That breaks SQL queries because the table doesn't exist. The first broken query results in a `wp_die()`, but that leads to more queries which in turn call `wp_die()`.

Passing `null` fixes the error while preserving the original intent from https://github.com/WordPress/wporg-parent-2021/pull/54#discussion_r1025741781.

I ran into this while trying to activate the parent theme on `events.wordpress.org` for https://github.com/WordPress/wordcamp.org/issues/1007


outdoor2kode (approved):
Thanks for the fix and the insight. Look good to me 👍

PR #101, opened 2023-08-31
Block styles: Update search block to match new theme designs

ryelle:
See https://github.com/WordPress/wporg-mu-plugins/issues/419 — the search field's styling is being updated in the next iteration of the redesign. This PR updates the current search styles to use this new design.

- Remove the "secondary style" since the old search styles are deprecated now (correct?)
- Adds a new style for "on dark" to handle the dark form
- The focus states are a little different than the mockup, since the icon-button and input are separate, both focusable elements. There's a video at the end of this to show what I mean. Happy to iterate on this.

Mockup:

Screenshot 2023-08-31 at 6 16 20 PM

### Screenshots

This will impact live sites using the parent theme, I think it's minimal (and expected, if this is the new search design that should be used). If that's not the case, we should figure out what cases use which search style. 

| Before | After |
|--------|-------|
| prod-showcase | search-showcase |
| prod-dev-blog | search-dev-blog |
| prod-docs | search-docs |
| prod-docs-single | search-docs-single |

The following video shows the various focus, hover, and filled-in states. I know most of these variations won't be used, but on the chance they are, it shouldn't break.

https://github.com/WordPress/wporg-parent-2021/assets/541093/462c97be-0f8b-4b30-bd41-16f75c9eb3b0

1. Use this branch for the parent theme, load any child theme project (showcase, developer, documentation, etc)
2. View any existing search blocks
3. Add search to a page, both styles & various configurations should work


source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
```diff
@@ -63,24 +74,31 @@
...
-			background: transparent;
-			background-image: url(../images/search-x.svg);
+			background-color: var(--wp--custom--form--active--color--text);
+			vertical-align: baseline;
+			mask-image: url(../images/search-x.svg?v=2);
+			mask-size: cover;
+		}
+
+		&:focus::-webkit-search-cancel-button {
+			background-color: var(--local--focus--outline-color);
```
adamwoodnz: For me on Developer, this changes the color of the clear icon to white when the input is focused and it disappears. From the screenshots I expected it to be blue, in which case I think this might need to be:

```suggestion
			background-color: var(--local--focus--border-color);
```

Developer search style

adamwoodnz (commented):
Great timing on this, thanks :)

Works for me on Developer. Just one odd thing I noticed, see comment inline.

jasmussen:
Nice!

source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
ryelle: Ah, thanks for catching that! I flipped those variable names to fix something else and missed the use here.

ryelle: Fixed in  https://github.com/WordPress/wporg-parent-2021/pull/101/commits/234b872425a8855d476dfadd8f9b4cc734bf289f

Screenshot 2023-09-01 at 12 29 32 PM

adamwoodnz (approved):
LGTM

PR #104, opened 2023-09-07
Templates: Add comments to single post template

### LINKED ISSUE #305: No comments in news?
I'm comparing in the ES version, 2 posts:

old: 
https://es.wordpress.org/2023/08/02/preocupacion-por-la-ley-de-ciberresiliencia-cra-de-la-union-europea/?new-theme=0

new:
https://es.wordpress.org/2023/08/02/preocupacion-por-la-ley-de-ciberresiliencia-cra-de-la-union-europea/?new-theme=1

And I cannot see the comments anywhere in the new version (I can see one comment in the old one). Are there yet?

ryelle:
The current templates do not show any comments, but the Rosetta blogs need to support commenting. This adds a simple comment layout to the end of single post views.

Fixes https://github.com/WordPress/wporg-main-2022/issues/305

This should not affect the other sites that are already using the parent theme — the developer blog has a custom template, while documentation, developer, & showcase override the `single.html` template. 

@iandunn FYI for events.wordpress.org, but I don't think you have any views that would use `single.html` anyway.

### Screenshots

@WordPress/meta-design I referenced the ".org website hub" figma for comment reference, but didn't find anything so I just tried to keep it simple.

| | Screenshot |
|---|---|
| Single comment, replies open |  |
| Multiple comments, replies closed (no form) |  |
| No comments yet, with comment form | Screen Shot 2023-09-07 at 16 59 14 |
| Threaded comments (rosetta sites don't support this currently) | Screenshot 2023-09-07 at 5 30 03 PM |

Try loading this PR on a sandbox and viewing the URL in the original issue.

To test locally…

1. Create & view a post with comments open
2. The reply form should appear
3. Add a comment
4. It should be loaded on the page
5. Add a few comments
6. They should show up as expected
7. View a post with comments disabled
8. There should be no reply form
9. Any previously entered comments will appear

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/templates/single.html
+++ b/source/wp-content/themes/wporg-parent-2021/templates/single.html
@@ -10,11 +10,11 @@
 	</section>
 	<!-- /wp:group -->
 
-	<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}}} -->
-	<div class="wp-block-group alignfull is-style-two-column-display" style="margin-top:var(--wp--preset--spacing--60)">
-		<!-- wp:comments -->
-		<div class="wp-block-comments">
-			<!-- wp:comments-title /-->
+	<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0"}}}} -->
+	<div class="wp-block-group alignfull is-style-two-column-display" style="margin-top:0">
+		<!-- wp:comments {"style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}}} -->
+		<div class="wp-block-comments" style="margin-top:var(--wp--preset--spacing--60)">
+			<!-- wp:comments-title {"style":{"spacing":{"margin":{"top":"0"}}}} /-->
 
 			<!-- wp:comment-template -->
 				<!-- wp:group {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}},"layout":{"type":"default"}} -->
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-parent-2021/templates/single.html
```diff
@@ -9,6 +9,52 @@
 		<!-- wp:post-content {"className":"is-style-two-column-display","layout":{"inherit":true}} /-->
 	</section>
 	<!-- /wp:group -->
+
+	<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}}} -->
+	<div class="wp-block-group alignfull is-style-two-column-display" style="margin-top:var(--wp--preset--spacing--60)">
```
StevenDufresne: If there are no comments, will this still render the margin?

StevenDufresne (approved):
Tested on my sandbox, looks good.

source/wp-content/themes/wporg-parent-2021/templates/single.html
ryelle: Yeah, it does, so there's a bigger gap here when there are no comments. I've moved the margin to the `comments` block and now it only shows up when there are comments or the comment form is open — https://github.com/WordPress/wporg-parent-2021/pull/104/commits/c97b7012a499ed0313f9391e5a7a3331ce311f31

| Before | After |
|---|---|
| Screen Shot 2023-09-08 at 16 19 02 | Screen Shot 2023-09-08 at 16 26 58 |

jasmussen:
Seems fine to me, we can follow up holistically if need be.

PR #112, opened 2023-09-29
Add styles for small, toggle and tertiary buttons

### LINKED ISSUE #111: Add styles for small, tertiary and toggle buttons
The filters and buttons on the Developer 'New and updated' page require new button styles.

The filters use small toggle buttons and a small tertiary button (Apply). The version selector uses a small button (Filter).

Screenshot 2023-09-29 at 5 04 52 PM

adamwoodnz:
Closes #111 

Adds 3 style variations:
- Toggle (See Developer design, states are based on [Design Library small button](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=2710%3A958&mode=dev))
- Tertiary (naming based on [Gutenberg storybook](https://wordpress.github.io/gutenberg/?path=/story/components-button--tertiary))
- Small (default with reduced padding, also see [Design Library small button](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=2710%3A958&mode=dev))

I've made Tertiary and Toggle small by definition, based on their semantics and usage in the Developer designs, and to avoid the complexity in supporting both large and small. We can add these variations if need be later.

#### States

| State | |
|--------|-------|
| Static | Screenshot 2023-10-02 at 3 40 43 PM |
| Hover | Screenshot 2023-10-02 at 3 41 08 PM |
| Focus | Screenshot 2023-10-02 at 3 41 37 PM |
| Active | Screenshot 2023-10-02 at 4 00 18 PM |
| Active Focus | Screenshot 2023-10-02 at 3 59 55 PM |

#### Developer

Screenshot 2023-10-02 at 4 08 28 PM

1. Test it with https://github.com/WordPress/wporg-developer/pull/292, the filters and version selector use all these styles
2. Create buttons in the editor and apply the style variations

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -81,14 +81,6 @@ function setup_block_styles() {
 	register_block_style(
 		'core/button',
 		array(
-			'name'         => 'small',
-			'label'        => __( 'Default Small', 'wporg' ),
-		)
-	);
-
-	register_block_style(
-		'core/button',
-		array(
 			'name'         => 'tertiary',
 			'label'        => __( 'Tertiary', 'wporg' ),
 		)
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -81,8 +81,8 @@ function setup_block_styles() {
 	register_block_style(
 		'core/button',
 		array(
-			'name'         => 'tertiary',
-			'label'        => __( 'Tertiary', 'wporg' ),
+			'name'         => 'text',
+			'label'        => __( 'text', 'wporg' ),
 		)
 	);
 
--- a/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/block-styles.php
@@ -79,6 +79,30 @@ function setup_block_styles() {
 	);
 
 	register_block_style(
+		'core/button',
+		array(
+			'name'         => 'small',
+			'label'        => __( 'Default Small', 'wporg' ),
+		)
+	);
+
+	register_block_style(
+		'core/button',
+		array(
+			'name'         => 'tertiary',
+			'label'        => __( 'Tertiary', 'wporg' ),
+		)
+	);
+
+	register_block_style(
+		'core/button',
+		array(
+			'name'         => 'toggle',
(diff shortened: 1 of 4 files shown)


jasmussen:
Looks good.

* I wonder if the hover on the toggle could be the same as on tertiary. Not blocking, and happy to defer to @fcoveram when he's back from AFK.
* I don't know that we need to increase the inner white halo on the active state on the small and default. I also wonder if the black for active may not be a bit overwhelming for at least toggle, and tertiary styles.

Small pet peeve, is it too late to rename "Tertiary" to "Text"? The former suggests a hierarchy where none exists. If we can't, that's fine.

ryelle:
I second the suggestion to rename "tertiary" to "text," that describes the button better too.

Perhaps a question for design, will we need small/toggle/tertiary variations on dark backgrounds? Or small outline buttons? Large text buttons? 

Since variations can't combine, we'd need to create potentially 16 individual variations: `[default, outline, text, toggle] * [small, normal] * [on dark, on light]` … and this picker already feels like a lot 😅  

Screenshot 2023-10-03 at 6 18 00 PM

If we might need more, maybe we could have toggles or something, like the `isSmall` prop on the GB button.

Screen Shot 2023-10-03 at 18 24 32

We can't go back on the "Default/Filled on dark" "Outline/Outline on dark" split, because those are already in use on production… but even just helper class that can be manually applied might help.

Screenshot 2023-10-03 at 6 13 54 PM

jasmussen:
> Perhaps a question for design, will we need small/toggle/tertiary variations on dark backgrounds? Or small outline buttons? Large text buttons?

I don't _think_ so. At the moment, [the Theme Directory](https://www.figma.com/file/DQuNUBmyVbbtJNOqCSrAPH/Themes?type=design&node-id=3229%3A4815&mode=design&t=O1CFXumXq6IwOX9T-1) is dark, but it's not using the toggles:

Screenshot 2023-10-04 at 09 39 20

Screenshot 2023-10-04 at 09 39 29

adamwoodnz:
> even just helper class that can be manually applied might help

Yeah I did consider whether this needed to be something exposed in the editor. If we go with the `is-small` helper class, should the `Toggle` and `Text` variations be defined large by default as well then?

adamwoodnz:
> should the `Toggle` and `Text` variations be defined large by default as well then?

I think so, otherwise the size difference would be unexpected. I've made this change now, so there are less variations:

Screenshot 2023-10-05 at 12 15 38 PM

Frontend:

PR #115, opened 2023-11-13
Add `wporg-parent-block-styles` as dependency of parent stylesheet

iandunn:
The block stylesheet is no longer implicitly enqueued after updating Core from `6.4-alpha-56702` to `6.5-alpha-57108`. That caused some styles on `wordpress.org/download/` and `developer.wordpress.org/redesign-test/` to disappear.

Related: `r21248-dotorg`
 
Props @adamwoodnz


adamwoodnz (approved):
Works for me :shipit:

PR #119, opened 2023-12-14
Enhance content wrapping with `text-wrap`

### LINKED ISSUE #118: Add text-wrap rules for handling widows and orphans
There are new CSS rules becoming available in browsers to handle widows and orphans, see https://developer.chrome.com/blog/css-text-wrap-pretty/

> The feature does a little more than just ensure paragraphs don't end with a single word, it also adjusts hyphenation if consecutive hyphenated lines appear at the end of a paragraph or adjusts previous lines to make room. It will also appropriately adjust for text justification. text-wrap: pretty is for generally better line wrapping and text breaking, currently focused on orphans. In the future, text-wrap: pretty may offer more improvements.

> There's also [text-wrap: balance](https://developer.chrome.com/blog/css-text-wrap-balance/), which doesn't prevent orphans, but does ensure the text wraps in a way that creates a harmonious text block. **I personally use balance for headlines and pretty for paragraphs.**

adamwoodnz:
Closes #118 

`text-wrap: balance` for headings and captions, results in harmonious text blocks.
`text-wrap: pretty` for longer form content, prevents orphans.

#### Before 
text wrap before

#### After
text wrap after

Test various sites with child themes of this theme ([Main](https://github.com/WordPress/wporg-main-2022), [Developer](https://github.com/WordPress/wporg-developer), [Showcase](https://github.com/WordPress/wporg-showcase-2022), [Documentation](https://github.com/WordPress/wporg-documentation-2022), [Events](https://github.com/WordPress/wordcamp.org/tree/production/public_html/wp-content/themes/wporg-events-2023)), ensuring your parent theme is on this branch and built.

Check the tags targeted, and see how they respond to different screen widths.

Check browsers that don't support the properties yet (Safari and Firefox at time of writing), they should gracefully degrade, ie. there should be no effect.
https://caniuse.com/?search=text-wrap%3Apretty
https://caniuse.com/?search=text-wrap%3Abalance

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -1,3 +1,23 @@
+// Create balanced text blocks
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+caption,
+figcaption {
+	text-wrap: var(--wp--custom--heading--typography--text-wrap);
+}
+
+// Prevent orphans
+p,
+ul,
+ol,
+blockquote {
+	text-wrap: var(--wp--custom--body--typography--text-wrap);
+}
+
 // Links
 a {
 	cursor: pointer;
--- a/source/wp-content/themes/wporg-parent-2021/theme.json
+++ b/source/wp-content/themes/wporg-parent-2021/theme.json
@@ -283,7 +283,8 @@
 			},
 			"body": {
 				"typography": {
-					"lineHeight": 1.875
+					"lineHeight": 1.875,
+					"textWrap": "pretty"
 				},
 				"short-text": {
 					"typography": {
@@ -328,7 +329,8 @@
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--eb-garamond)",
 					"fontWeight": 400,
-					"lineHeight": 1.3
+					"lineHeight": 1.3,
+					"textWrap": "balance"
 				},
 				"cta": {
 					"typography": {


source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
```diff
@@ -1,3 +1,23 @@
...
+h6,
+caption,
+figcaption {
+	text-wrap: var(--wp--custom--heading--typography--text-wrap);
+}
+
+// Prevent orphans
+p,
+ul,
+ol,
```
StevenDufresne, line 16: Does it need to be set on the `li` tag?

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
```diff
@@ -1,3 +1,23 @@
...
+h4,
+h5,
+h6,
+caption,
+figcaption {
+	text-wrap: var(--wp--custom--heading--typography--text-wrap);
+}
+
+// Prevent orphans
+p,
```
StevenDufresne, line 14: Do you think this will [cause performance issues](https://developer.chrome.com/blog/css-text-wrap-balance/#limitations) since we're targeting every `<p>` tag?

StevenDufresne: Thinking about the the [extra p tag](WordPress/wporg-developer/pull/430) problem on developer.

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
adamwoodnz: No, it gets inherited

source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
adamwoodnz: Yeah good point. Those limitations apply to `text-wrap: balance` which we're not using on p tags, only headings as [recommended](https://developer.chrome.com/blog/css-text-wrap-balance/#performance_considerations).

The [post](https://developer.chrome.com/blog/css-text-wrap-pretty) on `text-wrap: pretty` doesn't mention any limitations.

outdoor2kode: Saw someone has [tested the performance](https://twitter.com/csswizardry/status/1725543814587134435) of text-wrap: pretty; and text-wrap: balance; and found that thier overhead are quite similar, with balance being even a bit faster. Might be due to its limitation of only working for six wrapped lines and under. 
Overall the performance overhead for both is negligible.

Also saw this - applying balance to large paragraphs will have no effect and still comes at a performance cost because the browser is trying to calculate optimal balance even though it won't apply anything.

outdoor2kode (approved):
LGTM 👍

PR #120, opened 2024-01-15
Rosetta: Enable locale-specific customizations to theme.json settings.

### LINKED ISSUE #306: Homepage: Dream it section overflows in some languages
Origin:
https://wordpress.slack.com/archives/C02S78ZAL/p1691584951569959

At the new homepage if the "Dream it, build it" is longer than 250px, it breaks.

ryelle:
Add a method for overriding the default parent theme settings for rosetta sites. This will allow polyglots teams to provide, for example, preferred font sizes, which they or a meta dev can then code into the `get_locale_customizations` function using a `theme.json` format, so that any values [defined in settings](https://github.com/WordPress/wporg-parent-2021/blob/0b014ed6fe5b954e53afa046d9b278d1cd54463e/source/wp-content/themes/wporg-parent-2021/theme.json#L5-L682) can be adjusted. While the entire object can be updated (so that locales can also control the responsive sizes and line heights), some things like colors will likely not be allowed to change.

This PR works in tandem with https://github.com/WordPress/wporg-main-2022/pull/380, which adds a new CSS file for specific rosetta site overrides.

In practice, I see this working like so:

- Polyglots test out the theme with `https://[locale].wordpress.org/?new-theme=1`
- Notice some typography/alignment/spacing does not work well.
- Make an issue in wporg-main-2022 describing the issue.
- If the issue is page or section specific, a dev can make a PR to add CSS (like in https://github.com/WordPress/wporg-main-2022/pull/380).
- If the issue is more general, like universally font sizes should be smaller, the change should be made here to apply to all future sites as well.

If we end up with many customization requests, we could look at providing a UI for polyglots teams to manage this themselves, but I'd like to start here first.

❓ **Review question:** Is `get_locale()` the best thing to `switch` on, or is there a better, rosetta-specific function I should use?

See https://github.com/WordPress/wporg-main-2022/issues/266, Fixes https://github.com/WordPress/wporg-main-2022/issues/306

### Screenshots

I took a number of screenshots (with an automated script), you can [look through the whole batch here](https://cloudup.com/cy-tJdkZ-bB).

**Homepage**

In Catalan (and fr, it, ro), the heading CTA font is smaller (first two sections, community section, get started). In Japanese, the heading CTA and heading-2 fonts are smaller (all other headings).

| Locale | Before | After |
|-------|--------|-------|
| English (UK) | home-en-gb-before | home-en-gb-after |
| Catalan | home-ca-before | home-ca-after |
| Japanese | home-ja-before | home-ja-after |

**Download page**

Here, you can see the heading-2 size difference in the last two sections.

| Before | After |
|--------|-------|
| download-ja-before | download-ja-after |

You can test localized sites by [following the instructions on wporg-main-2022](https://github.com/WordPress/wporg-main-2022#working-on-non-english-sites).

I also tested this on my sandbox with the [top completed languages](https://translate.wordpress.org/projects/meta/wordpress-org/) for wporg.

After that, just clicking around pages and making sure they look okay. What I used:

- Home: '/'
- Download: '/download/'
- A download subpage: '/download/source/'
- About: '/about/'
- An about subpage: '/about/etiquette/'
- News: '/news/'

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/functions.php
+++ b/source/wp-content/themes/wporg-parent-2021/functions.php
@@ -8,6 +8,7 @@ defined( 'WPINC' ) || die();
 
 require_once __DIR__ . '/inc/gutenberg-tweaks.php';
 require_once __DIR__ . '/inc/block-styles.php';
+require_once __DIR__ . '/inc/rosetta-styles.php';
 
 /**
  * Actions and filters.
--- a/source/wp-content/themes/wporg-parent-2021/inc/rosetta-styles.php
+++ b/source/wp-content/themes/wporg-parent-2021/inc/rosetta-styles.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * Rosetta customizations.
+ */
+
+namespace WordPressdotorg\Theme\Parent_2021\Rosetta_Styles;
+
+defined( 'WPINC' ) || die();
+
+add_filter( 'wp_theme_json_data_user', __NAMESPACE__ . '\inject_i18n_customizations' );
+
+/**
+ * Inject customizations for Rosetta sites.
+ *
+ * @param WP_Theme_JSON_Data $theme_json Class to access and update the underlying data.
+ *
+ * @return WP_Theme_JSON_Data The updated user settings.
+ */
+function inject_i18n_customizations( $theme_json ) {
+	$locale_settings = get_locale_customizations( get_locale() );
+	if ( ! $locale_settings ) {
+		return $theme_json;
+	}
+
+	$config = array(
+		'version' => 2,
+		'settings' => $locale_settings,
+	);
+
+	return new \WP_Theme_JSON( $config, 'custom' );
+}
+
+/**
+ * Get a theme.json-shaped array with custom values for a given locale.
+ *
+ * The returned array should match the structure of "settings" in a theme.json
+ * file. These will be loaded as the "user" settings, which will override the
+ * theme.json values. Rosetta sites can then override any of the generated
+ * custom properties (ex, --wp--preset--font-size--normal) in a way that will
+ * cascade to any future child themes, and also render correctly in the editor.
+ *
+ * @param string $locale The current site locale.
+ *
+ * @return array An array of settings mirroring a theme.json "settings" object.
+ */
+function get_locale_customizations( $locale ) {
+	switch ( $locale ) {
+		case 'ca':
+		case 'fr':
+		case 'it_IT':
+		case 'ro_RO':
+			return [
+				'typography' => [
+					'fontSizes' => [
+						[
+							'slug' => 'heading-cta',
+							'size' => '96px',
+						],
+					],
+				],
+			];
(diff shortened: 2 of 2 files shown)


dd32 (approved):
As much as I don't like having to hard-code these, it makes sense to me as the route to take.

> ❓ Review question: Is get_locale() the best thing to switch on, or is there a better, rosetta-specific function I should use?

`get_locale()` is the best option here, Rosetta filters it on locale sites to the correct value.

t-hamano:
Thanks for the PR! There is one thing that concerns me about the Japanese site.

image

This line break is very strange to Japanese people.

For example, imagine the following displayed on an English site.

image

What we expect is to be displayed as follows.

image

We have also not inserted any special line break tags for this text.

image

I would also like to find out the cause, but do you know why this problem occurs?

ryelle:
> We have also not inserted any special line break tags for this text.

It looks like there are linebreaks in the strings, `<span>WordPress:</span><br />発信を<strong>すべての</strong>人の手に`

https://translate.wordpress.org/projects/meta/wordpress-org/ja/default/?filters%5Bterm%5D=%2Fsrc%2Frandom-heading%2Findex.php&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc

If I add CSS to make the bold text not wrap, that helps on the "Publish your passion" string. But I think the business string is still breaking in a bad place. If you remove those linebreaks though, I think the combination of both will fix it. How does the second column here look to you?

| | With linebreak | No linebreak |
|---|---|---|
| Passion | with-br-passion | no-br-passion |
| Business | with-br-business | no-br-business |
| Freedom | with-br-freedom | no-br-freedom |

t-hamano:
Thank you for your detailed testing.

> It looks like there are linebreaks in the strings, `<span>WordPress:</span><br />発信を<strong>すべての</strong>人の手に`

Sorry, I missed one br tag 😅😅

> How does the second column here look to you?

This depends on the variation being displayed, but in general, it looks more natural without a linebreak.

This PR simply changes the font size according to the locale, so this problem might be fixed by adding appropriate line breaks on GlotPress.

We will be able to find out the cause of this problem and fix it in advance by enabling preview mode. If a line break alone doesn't solve the problem, we might submit a PR adding some styles as a follow-up.

Sorry for confusing you 🙇

ryelle:
> This PR simply changes the font size according to the locale, so this problem might be fixed by adding appropriate line breaks on GlotPress.

Right, the line breaks need to be changed on GlotPress.

> If a line break alone doesn't solve the problem, we might submit a PR adding some styles as a follow-up.

That would be great.

> Sorry for confusing you 🙇

No confusion here — I just want to help you get a site that looks good :)

I'm going to merge this PR for the font size updates, and we can continue the conversation on https://github.com/WordPress/wporg-main-2022/issues/266

PR #124, opened 2024-02-14
Add three column layout helper class

adamwoodnz:
This PR ports the `.has-three-columns` helper class, used for layouts where sidebars float on each side of the content, from Developer Resources. Having it in this theme means that it can be reused for similar layouts in Documentation and likely other redesigned sites.

See https://github.com/WordPress/wporg-developer/issues/484

Try it in https://github.com/WordPress/wporg-developer/pull/499

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
@@ -26,3 +26,76 @@
 		max-width: var(--wp--custom--alignment--aligned-max-width);
 	}
 }
+
+// Configures a 3 column layout where `wporg/sidebar-container` blocks sit either side of the main content.
+// Examples: Developer Resources, Documentation.
+.has-three-columns {
+	--local--sidebar--width: 248px;
+	--local--column-gap: 40px;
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
```diff
@@ -26,3 +26,76 @@
 		max-width: var(--wp--custom--alignment--aligned-max-width);
 	}
 }
+
+// Configures a 3 column layout where `wporg/sidebar-container` blocks sit either side of the main content.
+// Examples: Developer Resources, Documentation.
+.has-three-columns {
```
adamwoodnz, line 32: Considered putting this in a new `blocks/_group.scss` partial and scoping it to `.wp-block-group`, but I'm not sure if there is any benefit.

ryelle: Keeping it in `_layout` makes sense to me 👍🏻

ryelle (approved):
Tested with https://github.com/WordPress/wporg-developer/pull/499, the 3-col layouts look good.

PR #125, opened 2024-02-15
Add custom code border radius

outdoor2kode:
As discussed in https://github.com/WordPress/wporg-developer/pull/447#discussion_r1461717206, a custom code block style is added for better semantics.

Please test with https://github.com/WordPress/wporg-developer/pull/447


adamwoodnz (approved):
👍 LGTM

PR #129, opened 2024-04-02
Change the color of visited links to purple

### LINKED ISSUE #128: Add different color for visited links.
Reported: https://meta.trac.wordpress.org/ticket/7529

The new theme doesn't have a color for visited links. I don't see it in the [figma](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=1040%3A3521&mode=dev).

 @WordPress/meta-design

adamwoodnz:
Changes the color of all links which are not buttons to a new purple color, added to the palette.

This has wide ranging effects and will need to be overridden in some cases. I've drafted [a corresponding mu-plugins PR](https://github.com/WordPress/wporg-mu-plugins/pull/595) for some of these, but it will probably need to be more expansive, see screenshots below.

Closes #128 
Related https://github.com/WordPress/wporg-mu-plugins/pull/595

### Screenshots (with local nav and breadcrumb overrides applied)

| Developer | Showcase | Forums |
|--------|-------|-|
| developer wordpress org_block-editor_getting-started_devenv_nodejs-development-environment_(2 1366x768) | wordpress org_showcase_(2 1366x768) | wordpress org_support_forum_installation_(Desktop) |

| Main | Documentation |
|--------|-------|
| wordpress org_(2 1366x768) | wordpress org_documentation_(Desktop) |

1. Build the styles
2. Test with dependent child themes; HelpHub, DevHub, Showcase, Main, etc
3. Visit links and observe the style afterward

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/theme.json
+++ b/source/wp-content/themes/wporg-parent-2021/theme.json
@@ -139,8 +139,18 @@
 				},
 				{
 					"slug": "purple-1",
-					"color": "#7a00df",
+					"color": "#5300be",
 					"name": "Purple 1"
+				},
+				{
+					"slug": "purple-2",
+					"color": "#7a00df",
+					"name": "Purple 2"
+				},
+				{
+					"slug": "purple-3",
+					"color": "#d7a7ff",
+					"name": "Purple 3"
 				}
 			]
 		},
--- a/source/wp-content/themes/wporg-parent-2021/theme.json
+++ b/source/wp-content/themes/wporg-parent-2021/theme.json
@@ -136,6 +136,11 @@
 					"slug": "lemon-3",
 					"color": "#fffdd6",
 					"name": "Lemon 3"
+				},
+				{
+					"slug": "purple-1",
+					"color": "#7a00df",
+					"name": "Purple 1"
 				}
 			]
 		},
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss
@@ -31,11 +31,11 @@ a {
 	&:focus {
 		text-decoration-line: none;
 	}
+}
 
-	&:where(:not(.wp-element-button)) {
-		&:visited {
-			color: var(--wp--preset--color--purple-1);
-		}
+:where(.entry-content a:not(.wp-element-button)) {
+	&:visited {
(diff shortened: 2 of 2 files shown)


StevenDufresne:
Does it make sense to style it differently using the background colors? 

**Examples**
`has-blueberry-1-background-color a:visited {}`
`has-charcoal-1-background-color a:visited {}`

jasmussen:
Perhaps a controversial take, but IMO we should not be showing visited states outside of the main body content. That is, anything in headers, footers, navigation sections, etc, those should be "evergreen links", insofar as they are meant to be navigational aids, fixtures, that never grow "visited". I'd love to start there, if we can, and it should solve the need outlined here. Notably the homepage stands out as something that should not have purple links here. We can revisit and add nuance if need be.

That said, here's a light version of the purple, for contexts that need it: #D7A7FF

ryelle:
I think this could work, since we are fairly consistent with applying the `entry-content` class to page contents:

```
:where(.entry-content a:not(.wp-element-button)) {
	&:visited {
		color: var(--wp--preset--color--purple-1);
	}
}
```

Using `:where` also keeps the specificity low enough so that sections & blocks using custom link colors stay that color.

Screenshot 2024-04-03 at 1 34 26 PM

Looks good on forums:

Screenshot 2024-04-03 at 1 36 16 PM

And documentation:

Screen Shot 2024-04-03 at 13 38 05

fcoveram:
The above looks great to me

adamwoodnz:
> I think this could work, since we are fairly consistent with applying the `entry-content` class to page contents

This is working nicely!

I've found places where we don't have the `.entry-content` class applied, eg. https://wordpress.org/documentation/category/installation/, but these are few and far between, and I think if that's our standard/best practice then this is acceptable.

source/wp-content/themes/wporg-parent-2021/theme.json
```diff
@@ -136,6 +136,21 @@
...
+				},
+				{
+					"slug": "purple-2",
+					"color": "#7a00df",
+					"name": "Purple 2"
+				},
+				{
+					"slug": "purple-3",
+					"color": "#d7a7ff",
+					"name": "Purple 3"
```
adamwoodnz, line 153: Palette defined [here](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=4706-96&mode=dev)

PR #131, opened 2024-04-05
Keep the same colors on hover.

StevenDufresne:
I find the hover state on the nav list distracting as it is almost a complete invert of the colors and isn't a necessary action as the item is already selected. This PRs removes the color flip on hover.

Thoughts?

# Screenshot
 
| Before | After |
|--------|--------|
| Screen Capture on 2024-04-05 at 15-17-41 | Screen Capture on 2024-04-05 at 15-21-25 |


ryelle:
It's still a link (and should stay a link, because it could be used to reset the filters), so it still needs some kind of hover indication. It could be different from the unselected hover, but it does need something.

Related, I also looked into whether the default hover state here is sufficient for WCAG and I don't think it is https://github.com/WordPress/wporg-parent-2021/issues/132

PR #133, opened 2024-04-05
Block styles: Add context to the translations for style names

ryelle:
The block style names are sent to GlotPress with no context, so it's hard to know how to translate them. This adds some content (simply, "block style name") to all the block styles. There should be no change for English sites.

@audrasjb Does this work, or should I only add it to the most-recently added names? IIRC this invalidates the previous translations, so I don't want to keep causing more work for you all 😅


audrasjb:
Thanks for handling this @ryelle!
I think the related translations will be fuzzied but I don't think it's an issue as long as it is communicated ahead of the change to polyglots via a Make/Polyglots post for big changes or a #polyglots announcement on Slack for small changes.

audrasjb (approved):

PR #137, opened 2024-05-08
Add cursor: pointer for `<button>`s.

StevenDufresne:
I'm working on [Accessibility: Plugin single links and buttons missing hover states #263](https://github.com/WordPress/wordpress.org/issues/263). 

That converted classic to block theme uses `<button>` and `<input>` elements in various places for form submission. 

In order to make those look the same as the new parent style, we've added `wp-block-button__link` to the elements and visual they are working.  👍 

The only difference is that we have `cursor: pointer` added to [anchor](https://github.com/WordPress/wporg-parent-2021/blob/7a7464cd6c3cf5b9ca87c09d8f68fa3182d944fd/source/wp-content/themes/wporg-parent-2021/sass/base/_elements.scss#L25) tags in this parent theme and therefore we get inconsistent hover states between `<a>` and `<button|input>` links.

This PR adds the `cursor: pointer` to the button as well. I can't think of a scenario where this wouldn't make sense.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_button-mixins.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_button-mixins.scss
@@ -4,6 +4,7 @@
 	@include button-typography-styles;
 	@include button-color-styles;
 	border-radius: var(--wp--custom--button--border--radius);
+	cursor: pointer;
 }
 
 @mixin button-color-styles {


ryelle (approved):
_Technically_, buttons are not meant to have pointer-cursors. The pointer is meant to denote links, specifically (this came up in conversation at the a11y meetup I went to on Monday, so it's funny timing).

However, we've been doing buttons with pointer cursors in core for at least 12 years, and most of the web does this too. And it's strange, when they can look identical, that two items would have different cursors. So, this does make them consistent 👍🏻

StevenDufresne:
That's great context and I appreciate the share. 

I personally find it weird that links are styled like buttons in modern web design as well. 

I'm also not opposed to figuring out a different visual indicator for actual buttons that are included in forms. Maybe there's a convention we can come up with form themselves in the future to denote that the button submits a form; be it color, outline, etc..

I'll merge this now to make the UI more consistent in the button/link sense but very open to future iterations.

Issue #139, opened 2024-06-07
Table of contents sidebar is too far down the page


ryelle:
When scrolled to the top of the page, the table of contents sidebar is too far offset.

![Screen Shot 2024-06-07 at 17 00 25](https://github.com/WordPress/wporg-developer/assets/541093/3126c1f0-3c59-45d6-a9e6-fc591d94af1a)

This seems to be new with the Gutenberg update, because constrained blocks now get `position:relative` https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/group/style.scss#L6

I'm not sure if we should unset the position:relative or adjust the top spacing on the sidebar.

adamwoodnz:
> This seems to be new with the Gutenberg update, because constrained blocks now get `position:relative` https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/group/style.scss#L6
> 
> I'm not sure if we should unset the position:relative or adjust the top spacing on the sidebar.

Agreed that this is the root cause. Transferring this to Parent where I think we should set the `.has-three-columns` class to `position: relative` so that we have more control over the layout from that parent node. This will require changes in the sidebar container block too, and possibly changes in the consuming themes.

ryelle:
@adamwoodnz Is there anything else to wrap up for this issue?

adamwoodnz:
Nope, we're done!

PR #140, opened 2024-06-10
Create positioning context for 3 column layout

adamwoodnz:
[Gutenberg 18.5 added `position:relative` to all constrained blocks](https://github.com/WordPress/gutenberg/pull/60347), breaking the 3 column layout.

> I'm not sure if we should unset the position:relative or adjust the top spacing on the sidebar.

I have gone for adjusting the top spacing, but also making the layout less brittle by defining it's own positioning context, and adjusting the position of the children accordingly. It has resulted in just as many changes across the site as if we had unset the relative position everywhere, but I think the end result is simpler, less brittle and moves Documentation forward in terms of redesign.

See https://github.com/WordPress/wporg-parent-2021/issues/139

Props @ryelle 

Probably easiest to load https://github.com/WordPress/wporg-developer/pull/522 and https://github.com/WordPress/wporg-documentation-2022/pull/97 in sandbox with this and https://github.com/WordPress/wporg-mu-plugins/pull/624

Test the sidebar container position on load and scrolling down. Test a page with a long ToC where it comes into contact with the footer, and check that the height of the sidebar is reduced to avoid overlap.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/base/_layout.scss
@@ -32,6 +32,7 @@
 .has-three-columns {
 	--local--sidebar--width: 248px;
 	--local--column-gap: 40px;
+	position: relative;
 	justify-content: flex-end;
 	max-width: 100%;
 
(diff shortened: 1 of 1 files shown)


ryelle (approved):
+1 for setting position:relative here, can guard against any future core changes.

Tested and good in developer & docs.

PR #151, opened 2024-08-28
Button styles: Add `is-destructive` helper class for red buttons

ryelle:
For some designs, we need a "destructive" red button, for example deleting a pattern or reporting a pledge. There isn't a consistent design for this, so for the Pattern Directory, the button uses a toggle style with a custom text color. This doesn't work for other buttons which have blue hover/focus styles— instead we need to create a helper class to set all the respective states.

Screenshot 2024-08-28 at 3 57 22 PM
Screenshot 2024-08-28 at 3 56 40 PM

This PR attempts to standardize on a few different reds— pomegrade-1 when contrasting with dark (text or backgrounds), `#d6310c` for the red on white, and `#b60000` as a slightly darker red for filled button hover states.

### Screenshots

Screenshot 2024-08-28 at 3 55 40 PM

Hover & focus styles:

https://github.com/user-attachments/assets/7d82eed5-f5f4-4bb4-8c06-a1b3fc16794f

**To test**

- Add some button blocks to a page
- In the button's sidebar, open Advanced and add the `is-destructive` class
- The button should be red now 🟥 
- Hover, focus, etc styles should also be red

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_button.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_button.scss
@@ -113,6 +113,50 @@
 		}
 	}
 
+	// Add "destructive" button styles.
+	&.is-destructive {
+		--wp--custom--button--color--text: var(--wp--preset--color--white);
+		--wp--custom--button--color--background: #d6310c;
+		--wp--custom--button--hover--color--background: #b60000;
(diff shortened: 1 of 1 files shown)


ryelle:
I'm assuming no feedback is a good to merge, since this is a minor blocker for Five for the Future. If there are any changes needed for this "destructive" button, create a new issue.

PR #156, opened 2024-09-16
Fix 404 page 'Oops!' text being cut off

adamwoodnz:
Fixes https://github.com/WordPress/wporg-parent-2021/issues/155

I'm not sure what has changed to make this happen, nor do I know what the original design was, so this is an attempt to stop the text being cropped both vertically and horizontally.

| Before 1536x768 |
|--------|
| wordpress org_duh(3 1536x864) (1) | 

### Changes

Adjusts the size of the text so that it fits horizontally, allows overflow to be visible, and adjusts layering so that the footer can overlap the text.

The `overflow: hidden` looks like it was added to stop the page scrolling due to the height of the 'Oops!' text, but I'm not observing that as being a problem. 

### Screenshots of most common screen sizes

| 1920x1080 | 1366x768 | 1536x864 |
|--------|-------|-|
| wordpress org_duh(1 1920x1080) | wordpress org_duh(2 1366x768) | wordpress org_duh(3 1536x864) |

| Tablet portrait | Tablet landscape | Mobile |
|-|-|-|
| wordpress org_duh(iPad) | wordpress org_duh(iPad) (1) | wordpress org_duh(Samsung Galaxy S20 Ultra) |

PR #160, opened 2024-09-25
Allow search box to be used

pkevan:
See: https://github.com/WordPress/wporg-parent-2021/issues/161 and https://github.com/WordPress/wporg-parent-2021/pull/156


ryelle:
The "homepage" link is also unclickable— what about putting a `z-index: -1` on the Oops container? That seems to help in my inspector testing.

pkevan:
Good catch - was only looking at the search box when testing!

PR #162, opened 2024-10-31
Remove the removal of the margin top.

StevenDufresne:
This relies on https://github.com/WordPress/wporg-mu-plugins/pull/665 and removes code that stops the output of the admin bar top margin.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/functions.php
+++ b/source/wp-content/themes/wporg-parent-2021/functions.php
@@ -54,9 +54,8 @@ function theme_support() {
 	// Experimental support for adding blocks inside nav menus
 	add_theme_support( 'block-nav-menus' );
 
-	// Remove the default margin-top added when the admin bar is used, this is
-	// handled by the theme, in `_site-header.scss`.
-	add_theme_support( 'admin-bar', array( 'callback' => '__return_false' ) );
+	// Add support for admin bar.
+	add_theme_support( 'admin-bar' );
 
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-parent-2021/functions.php
```diff
@@ -54,9 +54,8 @@ function theme_support() {
 	// Experimental support for adding blocks inside nav menus
 	add_theme_support( 'block-nav-menus' );
 
-	// Remove the default margin-top added when the admin bar is used, this is
-	// handled by the theme, in `_site-header.scss`.
-	add_theme_support( 'admin-bar', array( 'callback' => '__return_false' ) );
+	// Add support for admin bar.
+	add_theme_support( 'admin-bar' );
```
ryelle: This can be fully removed, admin-bar support is default. The only reason to add it is to set the `callback`.

ryelle: Also noting that this line came from when this was forked from the News theme, which is why it's referencing `_site-header.scss` — the parent theme never included that CSS, but the global header's CSS made it irrelevant. This same code was on News; but with the removal of the sticky header, I deleted the referenced CSS file (https://github.com/WordPress/wporg-news-2021/pull/431). In a follow-up commit https://github.com/WordPress/wporg-news-2021/commit/fb90c529c0480cba22798eca2f44345a39d4137a, I removed this code entirely.

PR #164, opened 2024-12-04
Theme.json: Add custom properties for Inter Heading font sizes

ryelle:
See #163 — This sets up the custom properties for "Inter Heading" font sizes and line heights. This also adds the same "breakpoint" switch as the EB Garamond headings, so using the custom property will automatically switch down to the "small screen" font size at 599px.

This can be used in the child themes like so:

```json
"styles": {
	"elements": {
		"h3": {
			"typography": {
				"fontSize": "var(--wp--custom--heading--level-5--inter--typography--font-size)",
				"lineHeight": "var(--wp--custom--heading--level-5--inter--typography--line-height)"
			}
		}
	}
}
```

### Screenshots

No visual change in this PR, this just makes the properties available.

Instructions in the child theme repo: https://github.com/WordPress/wporg-photo-directory/pull/37


ryelle:
cc @StevenDufresne @adamwoodnz — no review needed (unless you want to), just FYI that these properties exist now.

PR #165, opened 2024-12-09
Add a spinner style for use in interactive blocks.

StevenDufresne:
This PR converts the [javascript spinner](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/google-map/src/components/spinner.js) from the google-maps blocks to be used in child themes.

1. Add `<span class="wporg-spinner"></span>`


source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
...
+	display: block;
+	box-sizing: border-box;
+	animation: spinner-rotate 1.4s linear infinite;
+	transform-origin: center;
+	position: absolute;
+	top: 0;
+	left: 0;
+}
+
+@keyframes spinner-rotate {
```
adamwoodnz: Maybe namespace this too? Not sure how likely a collision is, but possible.
```suggestion
	animation: wporg-spinner-rotate 1.4s linear infinite;
	transform-origin: center;
	position: absolute;
	top: 0;
	left: 0;
}

@keyframes wporg-spinner-rotate {
```

adamwoodnz (dismissed):
Tried it out and the styles aren't being built. Need to add to `source/wp-content/themes/wporg-parent-2021/sass/blocks/_index.scss`

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
+.wporg-spinner {
+	--local-size: 16px;
+	--spinner-color: var(--wp--preset--color--blueberry-1, currentColor);
```
adamwoodnz: I misunderstood this, and tried to change the color of the spinning bit by putting the span in a paragraph block and changing the text color to orange, which didn't work. If we reversed this that would work, but would default to charcoal most of the time. Maybe that's ok, in allowing us to have editor customisation.

Use case would be on a page like SotW, where we show the map of watch parties and there is usually a custom link color.

Screenshot 2024-12-09 at 3 16 50 PM

StevenDufresne: > If we reversed this that would work, but would default to charcoal most of the time.

I don't know if that's possible seeing that currentColor would always be set or inherited so we would never get the blueberry coloring unless explicitly set. 

We could do something like this though:
```
--spinner-color: var(--wp--custom--wporg-spinner--color, var(--wp--preset--color--blueberry-1, currentColor));
```

We could then set `--wp--custom--wporg-spinner--color` to give it a specific color.

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
+.wporg-spinner {
```
ryelle: Minor, but can you move this file to `/base/`? `/blocks/*` should be for real blocks.

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
...
+	width: var(--local-size);
+	height: var(--local-size);
+	display: inline-block;
+	background-color: transparent;
+	position: relative;
+}
+
+/* Static outer circle */
+.wporg-spinner::before {
+	content: '';
```
ryelle: stylelint - this should use double-quotes
```suggestion
	content: "";
```

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
...
+	height: 100%;
+	border-radius: 50%;
+	border: 1.5px solid var(--background-stroke-color);
+	display: block;
+	box-sizing: border-box;
+}
+
+/* Rotating semicircle (arc) */
+.wporg-spinner::after {
+	content: '';
```
ryelle: stylelint - this should use double-quotes
```suggestion
	content: "";
```

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
...
+	position: absolute;
+	top: 0;
+	left: 0;
+}
+
+@keyframes wporg-spinner-rotate {
+	from {
+		transform: rotate(0deg);
+	}
+	to {
```
ryelle: stylelint - this needs an extra linebreak
```suggestion

	to {
```

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
ryelle: > --spinner-color: var(--wp--preset--color--blueberry-1, currentColor);

Since this CSS exists in the parent theme, `--wp--preset--color--blueberry-1` will always exist. So it will never use `currentColor`.

> Maybe that's ok, in allowing us to have editor customisation.

Do we need _editor_ customization? This seems like a helper class we might use in another block, where we could override with custom properties in the containing block. That said, I would change it to populate the custom properties from theme.json, for consistency with other block settings.

```json
"settings": {
	"custom": {
		"wporg-spinner": {
			"color": {
				"background": "var(--wp--preset--color--black-opacity-15)",
				"stroke": "var(--wp--preset--color--blueberry-1)"
			}
		}
```

Which would generate `--wp--custom--wporg-spinner--color--background` & `--wp--custom--wporg-spinner--color--stroke`, so you don't need to set them in the CSS either.

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
```diff
@@ -0,0 +1,49 @@
+.wporg-spinner {
+	--local-size: 16px;
```
ryelle: Does this need to be a custom property?

ryelle (commented):
In the past I've found it simple enough to just include the CSS for a spinner in the blocks using it ([example on patterns](https://github.com/WordPress/pattern-directory/blob/7208ff5fa48c0dd10ca2aad2d2b4e8a48f2d771f/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/style.scss#L47-L62), [example on themes](https://github.com/WordPress/wporg-theme-directory/blob/f1a9a8d75c7803eb18b37c2e4fdaf36068996bd9/source/wp-content/themes/wporg-themes-2024/src/theme-previewer/style.scss#L32-L51)). That's not a reason not to do this, but it is a _slightly_ different approach to the spinner.

Anyway, for this PR, I saw a few stylelint issues, I've noted them here. To run the stylelint checks locally, you can use `yarn workspace wporg-parent-2021 lint:css`

adamwoodnz:
> To run the stylelint checks locally, you can use `yarn workspace wporg-parent-2021 lint:css`

I noticed these using the stylelint extension for VSCode too

source/wp-content/themes/wporg-parent-2021/sass/blocks/_spinner.scss
adamwoodnz: In the [original](https://github.com/WordPress/wporg-mu-plugins/blob/37e457b27558dd11bc7875a8ab73629572e63e8f/mu-plugins/blocks/google-map/postcss/style.pcss#L78) I was using it for some other calculations and it's been copied over. I guess if width and height are expected to be the same value it still makes sense.

PR #169, opened 2024-12-19
Try setting the form border color to not be transparent

adamwoodnz:
When [setting up styles for the search box](https://github.com/WordPress/wporg-parent-2021/commit/4330ea36141a66705d6489f84ec8b6e59ca25d15#diff-fef21ca5fc98f62e9e60ad5dcaa4c2f4c4d206ff4ab42cefa234567ddbe92182), we set the form border color to be transparent, with 0 width. This has caused problems for inputs across the site, and we have added styles to child themes to overcome them.

See https://github.com/WordPress/wporg-make/issues/55, the latest report for the Make handbooks. Make has had temporary fixes added to handle this, but those could be removed with this PR merged.

This PR sets the default form border properties to the values we use across the site, and introduces a new custom property for the search border width.

### Screenshots

| Before | After |
|--------|-------|
| make wordpress org_photos_handbook_moderating-photos_(Desktop) (1) | make wordpress org_photos_handbook_moderating-photos_(Desktop) |

1. Build and load this branch in your sandbox or local environment
2. Check various pages where forms and search forms exist, eg. https://make.wordpress.org/photos/handbook/moderating-photos/, https://wordpress.org/plugins/developers/add/, https://wordpress.org/showcase/submit-a-wordpress-site/
3. The search form field wrapper should have no border, and the form input fields should have a border. You may need to toggle off child theme styles to see the parent styles in effect (definitely the case for Plugin directory, Make handbooks and Showcase).
4. Test focus states

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
+++ b/source/wp-content/themes/wporg-parent-2021/sass/blocks/_search.scss
@@ -46,7 +46,7 @@
 	&.wp-block-search__button-inside {
 		.wp-block-search__inside-wrapper {
 			padding: 0;
-			border-width: var(--wp--custom--form--border--width);
+			border-width: var(--wp--custom--form--search--border--width);
 			border-radius: var(--wp--custom--form--border--radius);
 			background-color: var(--wp--custom--form--search--color--background);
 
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Looks good to me, definitely +1 to unifying styles like this. I tested with the current themes and saw no regressions.

I think any style _removal_ in child themes would need to be separately checked/unified, for one example, the focus states on Showcase & Make form inputs (not search) are different, and should probably land on the Showcase style (1.5px blueberry). That might require another update here, if you wanted to add hover border & color properties.

PR #173, opened 2025-05-05
Fonts: Add Noto Serif SC for `zh_CN` locale, use filter for preloading fonts

### LINKED ISSUE #149: Add CJK subsets fonts support.
**Describe**
Hello, I'm the LM of Simplified Chinese (China). I found that the default fonts (EB Garamond and Latin) used in the new theme do not support CJK subsets, so I opened this issue (This isn't a big problem because browsers will automatically handle fallbacks).

For simplified Chinese, I suggest use [Source Han Sans (Noto Sans CJK)](https://github.com/adobe-fonts/source-han-sans), which is already used by Japanese websites and also supports Traditional Chinese and Korean.

Simplified Chinese (Singapore), Traditional Chinese (Hong Kong, Taiwan) and Korean rosetta site also need to be modified, but I am not sure about their LM's decision.

ryelle:
Fixes #149, see https://github.com/WordPress/wporg-mu-plugins/pull/684. This cleans up the font preloading slightly by moving the heading font preload call into the parent theme, and adding a filter around which font/subset is preloaded. This will help future updates by keeping the source of the font in `mu-plugins/global-fonts` and the loading of it in the parent theme's `rosetta-styles.php`. Initially the heading font was preloaded in each child theme to support different fonts per site, but that's not necessary and adds an extra step.

In addition to the filter & cleanup, this also loads Noto Serif SC for the Chinese (China) locale, and updates the Japanese locale to use Noto Serif JP anywhere EB Garamond had been.

Props @devhaozi.

### Screenshots

| Before | After |
|--------|-------|
| cn-eb-garamond | cn-noto-serif |

1. Load up this PR and https://github.com/WordPress/wporg-mu-plugins/pull/684
2. Set your site's language to Japanese or Chinese
3. View any page, it should preload Noto Serif SC or JP (view source to see)
4. Headings should be using the relevant Noto Serif.

### DIFF
--- a/source/wp-content/themes/wporg-parent-2021/functions.php
+++ b/source/wp-content/themes/wporg-parent-2021/functions.php
@@ -88,7 +88,8 @@ function enqueue_assets() {
 		/*
 		 * translators: Font subset for your locale. Can be any of cyrillic,
 		 * cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.
-		 * Do not translate into your own language.
+		 * Do not translate into your own language. If you don't use EB Garamond
+		 * for headings, you can ignore this.
 		 */
 		$subsets = _x( 'latin', 'EB Garamond subsets, comma separated', 'wporg' );
 		list( $font, $subsets ) = apply_filters( 'wporg_preload_heading_font', [ 'EB Garamond', $subsets ] );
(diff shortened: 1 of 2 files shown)


ryelle:
Noting that I've merged #174 down into this PR, so that it can be deployed along with https://github.com/WordPress/wporg-mu-plugins/pull/684.

I also added https://github.com/WordPress/wporg-parent-2021/pull/173/commits/7fc73efb7cc8f9d7836f0f517a2900b9522120ac — because we can't know which subset file to load, we need to disable preloading on locales with subset fonts.

